Issue #16, opened 2018-12-18
TypeError: __new__() missing 1 required positional argument: 'wait_for_ready'


rysloan4:
Hello,

I keep getting this grpc error and it's not entirely clear why:

`TypeError: __new__() missing 1 required positional argument: 'wait_for_ready'`

I'm running:
- Python 3.6.5
- google-ads==0.5.0

The code to reproduce is as follows:
```
google_ads_client = google.ads.google_ads.client.GoogleAdsClient.load_from_storage()
ga_service = google_ads_client.get_service('GoogleAdsService')
query = 'SELECT campaign.id, campaign.name FROM campaign ORDER BY campaign.id'
results = ga_service.search(customer_id, query=query,page_size=page_size)
for row in results:
     print(row)
```
The full stack trace is as follows:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/ryansloan/cupboard/py/reporting/env/lib/python3.6/site-packages/google/api_core/page_iterator.py", line 204, in _items_iter
    for page in self._page_iter(increment=False):
  File "/Users/ryansloan/cupboard/py/reporting/env/lib/python3.6/site-packages/google/api_core/page_iterator.py", line 235, in _page_iter
    page = self._next_page()
  File "/Users/ryansloan/cupboard/py/reporting/env/lib/python3.6/site-packages/google/api_core/page_iterator.py", line 526, in _next_page
    response = self._method(self._request)
  File "/Users/ryansloan/cupboard/py/reporting/env/lib/python3.6/site-packages/google/api_core/gapic_v1/method.py", line 143, in __call__
    return wrapped_func(*args, **kwargs)
  File "/Users/ryansloan/cupboard/py/reporting/env/lib/python3.6/site-packages/google/api_core/retry.py", line 270, in retry_wrapped_func
    on_error=on_error,
  File "/Users/ryansloan/cupboard/py/reporting/env/lib/python3.6/site-packages/google/api_core/retry.py", line 179, in retry_target
    return target()
  File "/Users/ryansloan/cupboard/py/reporting/env/lib/python3.6/site-packages/google/api_core/timeout.py", line 214, in func_with_timeout
    return func(*args, **kwargs)
  File "/Users/ryansloan/cupboard/py/reporting/env/lib/python3.6/site-packages/google/api_core/grpc_helpers.py", line 57, in error_remapped_callable
    return callable_(*args, **kwargs)
[...]

PR #17, opened 2018-12-19
Fix bug related to augmenting client call details metatadata

BenRKarl:
Addresses this issue: https://github.com/googleads/google-ads-python/issues/16

Implementation based [on this](https://github.com/grpc/grpc/blob/master/examples/python/interceptors/headers/header_manipulator_client_interceptor.py) example from grpc.

### DIFF
--- a/google/ads/google_ads/client.py
+++ b/google/ads/google_ads/client.py
@@ -16,7 +16,7 @@
 import logging
 import os
 import yaml
-import collections
+from collections import namedtuple
 
 import google.api_core.grpc_helpers
 import google.auth.transport.requests
@@ -378,7 +378,7 @@ class MetadataInterceptor(grpc.UnaryUnaryClientInterceptor):
 
 
 class _ClientCallDetails(
-        collections.namedtuple(
+        namedtuple(
             '_ClientCallDetails',
             ('method', 'timeout', 'metadata', 'credentials')),
         grpc.ClientCallDetails):
--- a/google/ads/google_ads/client.py
+++ b/google/ads/google_ads/client.py
@@ -16,6 +16,7 @@
 import logging
 import os
 import yaml
+import collections
 
 import google.api_core.grpc_helpers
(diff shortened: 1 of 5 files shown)


google/ads/google_ads/client.py
```diff
@@ -346,6 +347,21 @@ def __init__(self, developer_token, login_customer_id):
             ('login-customer-id', login_customer_id) if login_customer_id
             else None)
 
+    def _update_client_call_details_metadata(
+            self, client_call_details, metadata):
+        class _ClientCallDetails(
```
ericgribkoff: Google style guide recommends avoiding inner classes - I would move this definition outside of `class MetadataInterceptor`

BenRKarl: Awesome, that's great feedback. That would also make it easier to access in the tests if we decided to do so.

google/ads/google_ads/client.py
```diff
@@ -16,6 +16,7 @@
 import logging
 import os
 import yaml
+import collections
```
wihl: suggest `from collections import namedtuple`

google/ads/google_ads/client.py
```diff
@@ -361,14 +370,22 @@ def intercept_unary_unary(self, continuation, client_call_details, request):
...
-        )
+        client_call_details = self._update_client_call_details_metadata(
+            client_call_details,
+            metadata)
 
         return continuation(client_call_details, request)
 
 
+class _ClientCallDetails(
+        collections.namedtuple(
```
wihl: then this would be `namedtuple`

google/ads/google_ads/client.py
BenRKarl: :+1:

google/ads/google_ads/client.py
BenRKarl: :+1:

PR #18, opened 2018-12-20
Adding account_management/list_accessible_customers example

BenRKarl:


examples/v0/account_management/get_accessible_customers.py
```diff
@@ -0,0 +1,52 @@
...
+        print('Request with ID "%s" failed with status "%s" and includes the '
+              'following errors:' % (ex.request_id, ex.error.code().name))
+        for error in ex.failure.errors:
+            print('\tError with message "%s".' % error.message)
+            if error.location:
+                for field_path_element in error.location.field_path_elements:
+                    print('\t\tOn field: %s' % field_path_element.field_name)
+        sys.exit(1)
+
+    resource_names = accessible_customers.resource_names
```
wihl: suggest you add this and the iteration into the try: block

examples/v0/account_management/get_accessible_customers.py
```diff
@@ -0,0 +1,52 @@
...
+
+    for resource_name in resource_names:
```
wihl: no need for a one-off throw away name, ie. `for resource_name in accessible_customers.resources_names:`

examples/v0/account_management/get_accessible_customers.py
```diff
@@ -0,0 +1,52 @@
...
+        print('Customer account resource name: %s' % resource_name)
```
wihl: some customers have thousands of accounts. I suggest you put in a max limit to clip the output after say 200 items.

examples/v0/account_management/get_accessible_customers.py
BenRKarl: Great, I did keep this in order to slice the result to 200 entries, per your other comment.

examples/v0/account_management/get_accessible_customers.py
BenRKarl: I added a line to slice the result array down to 200 entries.

examples/v0/account_management/get_accessible_customers.py
```diff
@@ -0,0 +1,52 @@
...
+
+import google.ads.google_ads.client
+
+
+def main(client):
+    customer_service = client.get_service('CustomerService')
+
+    try:
+        accessible_customers = customer_service.list_accessible_customers()
+        resource_names = accessible_customers.resource_names[:200]
```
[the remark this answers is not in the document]
jradcliff: +1 to David's suggestion. See also my comment below.

examples/v0/account_management/get_accessible_customers.py
```diff
@@ -0,0 +1,52 @@
...
+        for resource_name in resource_names:
+            print('Customer account resource name: %s' % resource_name)
```
jradcliff: Print a message after the for loop if len(accessible_customers.resource_names) > 200 so users know the list was truncated.

examples/v0/account_management/get_accessible_customers.py
jradcliff: Instead of using a literal here, make the limit an argument that defaults to 200. That way, a user who really wants thousands of customers back can get them without modifying the example.

examples/v0/account_management/get_accessible_customers.py
jradcliff: That message should indicate the limit + how many accessible customers there are in total

jradcliff: nit: "account" is redundant here

examples/v0/account_management/get_accessible_customers.py
BenRKarl: Done.

examples/v0/account_management/get_accessible_customers.py
BenRKarl: :+1:

examples/v0/account_management/get_accessible_customers.py
BenRKarl: Good idea

examples/v0/account_management/get_accessible_customers.py
BenRKarl: I added two messages - one that prints the total before the loop and a second that notifies of a truncation if the total number is greater than the limit.

BenRKarl: Fixed here and at the top of the file.

examples/v0/account_management/get_accessible_customers.py
```diff
@@ -0,0 +1,68 @@
...
+
+from __future__ import absolute_import
+
+import argparse
+import six
+import sys
+
+import google.ads.google_ads.client
+
+_DEFAULT_LIMIT=200
```
AnashOommen: Why is this limit required? Google Ads limit this to 20 accounts anyway.

https://support.google.com/google-ads/answer/6139186?hl=en

BenRKarl: It was a suggestion from @wihl and @jradcliff for the larger clients that may have thousands of customers. It just restricts the number of results that are logged, it doesn't affect the actual request. Also those docs say that up to 20 customers can be associated with a single email address, but does that also apply to customers that are connected to a manager account? I thought those were different concepts. 

I was curious if the API itself would restrict the number of results returned here, but I couldn't find any documentation to support that. Also there doesn't appear to be a way to paginate these results. We figured a limit of 200 was a reasonable number to keep logs manageable.

BenRKarl:
@AnashOommen Added a comment to the documentation at the top of the file with details on how to use the IDs retrieved by this request. Thanks for the suggestion!

examples/v0/account_management/get_accessible_customers.py
jradcliff: @AnashOommen raises an excellent point that I failed to recognize: the request will only return customers that are *directly* accessible by the user, and that is indeed limited to 20 customers. +1 to removing this parameter and limit, and sorry I didn't think of this earlier.

BenRKarl: @jradcliff @AnashOommen No problem! I'll strip it out now.

examples/v0/account_management/get_accessible_customers.py
```diff
@@ -63,10 +58,5 @@ def main(client, limit=_DEFAULT_LIMIT):
     parser = argparse.ArgumentParser(
         description=('Lists all customers the authenticating user has '
                      'access to.'))
```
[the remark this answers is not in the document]
BenRKarl: Ah good catch, no it is not needed. I removed it.

examples/v0/account_management/get_accessible_customers.py
```diff
@@ -0,0 +1,59 @@
...
+authenticating user has access to.
+
+The customer IDs retrieved from the resource names can be used to set
+the login-customer-id configuration. For more information see this
+documentation: https://developers.google.com/google-ads/api/docs/concepts/call-structure#login-customer-id
+"""
+
+from __future__ import absolute_import
+
+import argparse
```
jradcliff: Do you still need to import `argparse` and `six`?

BenRKarl: Ack - nope - thank you for catching.

PR #19, opened 2018-12-20
Adding account_management/create_customer.py

BenRKarl:


examples/v0/account_management/create_customer.py
```diff
@@ -0,0 +1,71 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""This example illustrates how to create a new customer under a given
+manager account.
+
+Note by default this account will only be accessible via its parent AdWords
```
wihl: Check the branding - I don't think we call this AdWords anymore

examples/v0/account_management/create_customer.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def main(client, customer_id):
+    customer_service = client.get_service('CustomerService')
+    resource_name = customer_service.customer_path(customer_id)
+    customer = client.get_type('Customer')
+    customer.resource_name = resource_name
+    customer.descriptive_name.value = "A new customer."
+    customer.currency_code.value = "USD"
+    customer.time_zone.value = "America/New_York"
```
wihl: Suggest a link to https://developers.google.com/adwords/api/docs/appendix/codes-formats for the list of Timezone IDs

examples/v0/account_management/create_customer.py
```diff
@@ -0,0 +1,71 @@
...
+    customer.tracking_url_template.value = "{lpurl}?device={device}"
+    customer.final_url_suffix.value = "param1=value1&param2=value2"
```
wihl: You can probably improve the final URL example to be more meaningful for conversions. Also, I believe this string should be URL encoded. In Python, urllib can do this.

jradcliff: For a more meaningful value, you could use something like this for the final url suffix:

    keyword={keyword}&matchtype={matchtype}&adgroupid={adgroupid}

where the params above come from https://support.google.com/google-ads/answer/6305348.

You do not need to encode either of these strings. AdWords will properly encode any values from ValueTrack parameters. Of course, if you had a literal here, you'd want to encode it properly, e.g.,

    fact=ben%20is%the%20greatest

There's a bit more info on all of this in the HC article above.

examples/v0/account_management/create_customer.py
BenRKarl: Good catch!

examples/v0/account_management/create_customer.py
BenRKarl: I think that's a great idea, I'll add a comment for it.

examples/v0/account_management/create_customer.py
```diff
@@ -14,7 +14,7 @@
 """This example illustrates how to create a new customer under a given
```
msaniscalchi: *nitpick*: 1-liner summary should be 1 line.

examples/v0/account_management/create_customer.py
```diff
@@ -32,11 +32,14 @@ def main(client, customer_id):
     resource_name = customer_service.customer_path(customer_id)
```
msaniscalchi: It looks like indentation is wrong starting here through the try block (interestingly... the except block is O.K.)

examples/v0/account_management/create_customer.py
```diff
@@ -0,0 +1,71 @@
...
+    customer.descriptive_name.value = "A new customer."
+    customer.currency_code.value = "USD"
+    customer.time_zone.value = "America/New_York"
+    customer.tracking_url_template.value = "{lpurl}?device={device}"
+    customer.final_url_suffix.value = "param1=value1&param2=value2"
+    customer.has_partners_badge.value = False
+
+    try:
+      response = customer_service.create_customer_client(
+        customer_id, customer)
```
msaniscalchi: 4 spaces indentation.

examples/v0/account_management/create_customer.py
```diff
@@ -0,0 +1,71 @@
...
+                         .load_from_storage())
+
+    parser = argparse.ArgumentParser(
+        description=('Creates a new client under the given manager.'))
+    # The following argument(s) should be provided to run the example.
+    parser.add_argument('-c', '--customer_id', type=six.text_type,
+                        required=True, help='The Google Ads customer ID.')
+    args = parser.parse_args()
+
+    main(google_ads_client, args.customer_id)
```
msaniscalchi: *nitpick*:
Should end on a newline.

examples/v0/account_management/create_customer.py
BenRKarl: I'm not sure I follow. This is formatted similar to comments in other examples, see https://github.com/googleads/google-ads-python/blob/master/examples/v0/account_management/get_account_information.py#L14

examples/v0/account_management/create_customer.py
BenRKarl: Is this not indented by 4 spaces? It looks correct to me.

examples/v0/account_management/create_customer.py
BenRKarl: Good catch, fixed this so it's indented by 4 not 2.

examples/v0/account_management/create_customer.py
BenRKarl: The file does end with a newline, strangely it just doesn't seem to be displayed. I think GitHub may not render them - they don't appear in other files.

examples/v0/account_management/create_customer.py
```diff
@@ -0,0 +1,76 @@
...
+
+import google.ads.google_ads.client
+
+
+def main(client, customer_id):
+    customer_service = client.get_service('CustomerService')
+    resource_name = customer_service.customer_path(customer_id)
+    customer = client.get_type('Customer')
+    customer.resource_name = resource_name
+    customer.descriptive_name.value = 'A new customer.'
```
jradcliff: Although the descriptive name does not need to be unique, let's use the [same approach we used with the SOAP example](https://github.com/googleads/googleads-python-lib/blob/master/examples/adwords/v201809/account_management/create_account.py#L44) so users can easily find the new account.

```
today = datetime.today().strftime('%Y%m%d %H:%M:%S')
customer.descriptive_name.value = 'Account created with CustomerService on %s' % today
```

examples/v0/account_management/create_customer.py
```diff
@@ -0,0 +1,76 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""This example illustrates how to create a new customer under a given
+manager account.
+
+Note by default this account will only be accessible via its parent Google Ads
+manager account.
```
jradcliff: Please use the note from the [Java SOAP example](https://github.com/googleads/googleads-java-lib/blob/master/examples/adwords_axis/src/main/java/adwords/axis/v201809/accountmanagement/CreateAccount.java#L41), with a few minor tweaks:

```
Note: this example must be run using the credentials of a Google Ads manager account. By
default, the new account will only be accessible via the manager account.
```

examples/v0/account_management/create_customer.py
```diff
@@ -0,0 +1,76 @@
...
+    customer.tracking_url_template.value = '{lpurl}?device={device}'
+    customer.final_url_suffix.value = ('keyword={keyword}&matchtype={matchtype}'
+                                       '&adgroupid={adgroupid}')
+    customer.has_partners_badge.value = False
+
+    try:
+      response = customer_service.create_customer_client(
+        customer_id, customer)
+      print(('Customer created with resource name "%s" under manager account'
+             ' with CID %s') % (response.resource_name, customer_id))
```
jradcliff: `CID` -> `customer ID`

examples/v0/account_management/create_customer.py
```diff
@@ -0,0 +1,76 @@
...
+    resource_name = customer_service.customer_path(customer_id)
+    customer = client.get_type('Customer')
+    customer.resource_name = resource_name
+    customer.descriptive_name.value = 'A new customer.'
+    # For a list of valid currency codes and time zones see this documentation:
+    # https://developers.google.com/adwords/api/docs/appendix/codes-formats
+    customer.currency_code.value = 'USD'
+    customer.time_zone.value = 'America/New_York'
+    # The below values are optional. For more information about tracking url
+    # templates see: https://support.google.com/google-ads/answer/6305348
```
jradcliff: nit: the article also covers final URL suffix, final URLs, etc., so instead of "tracking url templates" please use "URL options".

examples/v0/account_management/create_customer.py
msaniscalchi: Summary line appears > 80 chars.

examples/v0/account_management/create_customer.py
BenRKarl: Done

examples/v0/account_management/create_customer.py
BenRKarl: Done

examples/v0/account_management/create_customer.py
BenRKarl: Done

examples/v0/account_management/create_customer.py
BenRKarl: Done

examples/v0/account_management/create_customer.py
msaniscalchi: I'm seeing an indent of six spaces over here for `resource_name`. Is that not the case?

BenRKarl: I think that's just formatting applied by GitHub. If you click "view file" above it'll display the unformatted code and you should only see 4 spaces.

examples/v0/account_management/create_customer.py
BenRKarl: Weird, this line isn't over 80 chars - I wonder if it's just silliness with out GitHub is displaying the diff.

examples/v0/account_management/create_customer.py
```diff
@@ -0,0 +1,80 @@
...
+if __name__ == '__main__':
+    # GoogleAdsClient will read the google-ads.yaml configuration file in the
+    # home directory if none is specified.
+    google_ads_client = (google.ads.google_ads.client.GoogleAdsClient
+                         .load_from_storage())
+
+    parser = argparse.ArgumentParser(
+        description=('Creates a new client under the given manager.'))
+    # The following argument(s) should be provided to run the example.
+    parser.add_argument('-c', '--customer_id', type=six.text_type,
```
fiboknacky: Can we use different argument name from the other examples, like manager customer ID, to make it explicit?

examples/v0/account_management/create_customer.py
```diff
@@ -0,0 +1,80 @@
...
+import six
+import sys
+from datetime import datetime
+
+import google.ads.google_ads.client
+
+
+def main(client, customer_id):
+    customer_service = client.get_service('CustomerService')
+    resource_name = customer_service.customer_path(customer_id)
```
fiboknacky: This is unnecessary and actually ignored by the server.
It's confusing to have this here too, as the specified parameter should be "manager customer ID", and that should be specified when you call "customer_service.create_customer_client" only.

This object is a new one so it shouldn't have existing customer ID nor resource name yet.

examples/v0/account_management/create_customer.py
BenRKarl: Sure!

examples/v0/account_management/create_customer.py
BenRKarl: Great point, I removed this line

examples/v0/account_management/create_customer.py
```diff
@@ -0,0 +1,78 @@
...
+from datetime import datetime
+
+import google.ads.google_ads.client
+
+
+def main(client, manager_customer_id):
+    customer_service = client.get_service('CustomerService')
+    customer = client.get_type('Customer')
+    today = datetime.today().strftime('%Y%m%d %H:%M:%S')
+    customer.descriptive_name.value = ('Account create with '
```
msaniscalchi: *nitpick* 
"create" --> "created"

examples/v0/account_management/create_customer.py
```diff
@@ -0,0 +1,78 @@
...
+    customer.tracking_url_template.value = '{lpurl}?device={device}'
+    customer.final_url_suffix.value = ('keyword={keyword}&matchtype={matchtype}'
+                                       '&adgroupid={adgroupid}')
+    customer.has_partners_badge.value = False
+
+    try:
+        response = customer_service.create_customer_client(
+            manager_customer_id, customer)
+        print(('Customer created with resource name "%s" under manager account'
+               ' with customer ID %s') % (response.resource_name, manager_customer_id))
```
msaniscalchi: *nitpick*:
When using implicit string concatenation like this, avoid starting lines with whitespace--e.g. the previous line should end with whitespace.

examples/v0/account_management/create_customer.py
```diff
@@ -0,0 +1,78 @@
...
+    # GoogleAdsClient will read the google-ads.yaml configuration file in the
+    # home directory if none is specified.
+    google_ads_client = (google.ads.google_ads.client.GoogleAdsClient
+                         .load_from_storage())
+
+    parser = argparse.ArgumentParser(
+        description=('Creates a new client under the given manager.'))
+    # The following argument(s) should be provided to run the example.
+    parser.add_argument('-m', '--manager_customer_id', type=six.text_type,
+                        required=True, help='The Google Ads customer ID.')
```
msaniscalchi: If we're going to change the argument name, we probably want to be more descriptive in the help text.

examples/v0/account_management/create_customer.py
BenRKarl: Good to know! I'll keep that in mind from now on.

examples/v0/account_management/create_customer.py
BenRKarl: Yes great point, updated!

examples/v0/account_management/create_customer.py
```diff
@@ -0,0 +1,81 @@
...
+    customer.tracking_url_template.value = '{lpurl}?device={device}'
+    customer.final_url_suffix.value = ('keyword={keyword}&matchtype={matchtype}'
+                                       '&adgroupid={adgroupid}')
+    customer.has_partners_badge.value = False
+
+    try:
+        response = customer_service.create_customer_client(
+            manager_customer_id, customer)
+        print(('Customer created with resource name "%s" under manager account '
+               'with customer ID %s') %
```
msaniscalchi: **nitpick**: Resource names and IDs are usually wrapped in quotations.

BenRKarl: Ah so like `'with customer ID "%s"'`? I'll update here, and in the other PR too if needed.

PR #21, opened 2019-01-08
Updating PyYAML dependency due to security issue

wihl:
PyYAML 4.2 is experimental, so install with `pip install PyYAML==4.2b4`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ with io.open('README.rst', 'r', encoding='utf-8') as readme_file:
 
 setup(
     name='google-ads',
-    version='0.5.1',
+    version='0.5.2',
     author='Google LLC',
     author_email='googleapis-packages@google.com',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -20,7 +20,7 @@
     'enum34; python_version < "3.4"',
     'google-auth-oauthlib>=0.0.1,<1.0.0',
     'google-api-core[grpc] >= 1.1.0, < 2.0.0dev',
-    'PyYAML >= 3.11, < 4.0',
+    'PyYAML >=4.2b1',
```
msaniscalchi: As an FYI, a lack of an upper-bound has burned us before. :-)

Generally speaking, a major version bump is probably a good point to work with, as those are expected to include breaking changes.

wihl: Good idea - added.

Issue #24, opened 2019-01-23
Exception not being parsed properly


MarTango:
Related to #8

There still seems to be an issue, with `AttributeError: '_Rendezvous' object has no attribute 'request_id'` on [this line](https://github.com/googleads/google-ads-python/blob/master/google/ads/google_ads/client.py#L363). This can easily be replicated with a `GoogleAdsService.search` request with an syntactically invalid query.

BenRKarl:
Hi @MarTango - 

Thanks for bringing this up! Indeed it appears to be a bug of some kind. We're investigating and should post a resolution or bug fix soon.

BenRKarl:
@MarTango can you clarify which version of Python you're using when you encounter this error?

MarTango:
3.7.2

On 23 Jan 2019 17:52, "Ben Karl" <notifications@github.com> wrote:

@MarTango <https://github.com/MarTango> can you clarify which version of
Python you're using when you encounter this error?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/googleads/google-ads-python/issues/24#issuecomment-456901152>,
or mute the thread

BenRKarl:
@MarTango See the above PR that links to this issue for context on the bug.

Temporarily you can work around the issue by manually installing `grpcio==1.17.0` into your environment after installing `google-ads`, specifically (if you're using `pip`):

```
$ pip install google-ads
$ pip uninstall grpcio
$ pip install grpcio==1.17.0
```

MarTango:
Good news!
Thanks!

(Github isn't allowing me to comment, so I'm replying to the email instead)

MarTango:
Unfortunately that temporary solution doesn't work for me.

I get a `TypeError: Object of type bytes is not JSON serializable`

BenRKarl:
@MarTango Thanks for sharing this info. I suspect it's related to the issue I described in the PR. Pinning those dependency versions is a short-term stopgap; we're working on a more complete fix that should hopefully address the larger issue.

Can you tell me what type of request you're making to generate this error?

MarTango:
I think any invalid GAQL query will do it: e.g. `next(google_ads_service.search(customer_id, "SELECT campaign.idz FROM campaign").__iter__())`

BenRKarl:
[...]

PR #25, opened 2019-01-23
Separate google-api-core and grpcio deps in order to pin grpcio

BenRKarl:
Related to this issue: https://github.com/googleads/google-ads-python/issues/24

We've been using the `google-api-core[grpc]` dependency, which contains the extra (`[grpc]`) that, along with all of `google-api-core` installs `grpcio >= 1.8.2` See [here](https://github.com/googleapis/google-cloud-python/blob/master/api_core/setup.py#L42).

Recently `grpcio` version 1.18.0 was released and implicitly installed in our library as a result of this extra. Unfortunately the new release changed the error object structure returned in our interceptors [here](https://github.com/googleads/google-ads-python/blob/master/google/ads/google_ads/client.py#L510), generating new errors. 

This fix separates the two dependencies and pins `grpcio` to 1.17.0 to stabilize things while we determine the best long-term solution.

EDIT:

Some additional context - [this change](https://github.com/grpc/grpc/commit/8199aff7a66460fbc4e9a82ade2e95ef076fd8f9#diff-b98ef07fc6538e04ffb4d0287b6e34ac) appears to be the root of our problems. Notice in our code we use a try/catch and convert the error if it's a certain class, which is then more easily parsed in the logging interceptor. Since it isn't raised anymore the friendly errors don't get converted and the logging interceptor is unable to process them. Shouldn't be difficult to address this at the root, but I do think we should still pin our dependencies and upgrade them by hand for now.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,10 +18,10 @@ import io
 
 install_requires = [
     'enum34; python_version < "3.4"',
-    'google-auth-oauthlib>=0.0.1,<1.0.0',
-    'google-api-core >= 1.1.0, < 2.0.0dev',
+    'google-auth-oauthlib >= 0.0.1, < 1.0.0',
+    'google-api-core == 1.7.0',
     'grpcio == 1.17.0',
-    'PyYAML >=4.2b1, < 5.0',
+    'PyYAML >= 4.2b1, < 5.0',
 ]
 
 tests_require = [
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,8 @@ import io
 install_requires = [
     'enum34; python_version < "3.4"',
     'google-auth-oauthlib>=0.0.1,<1.0.0',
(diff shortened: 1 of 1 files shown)


wihl:
Ben and I discussed this and came to the conclusion that this is the best course of action given the imminent release of Google Ads API V1.0

BenRKarl:
@msaniscalchi I have run a few examples to verify things still work. Also even though `grpcio` has bumped from 1.17.0 to 1.18.0 recently, `google-api-core` hasn't updated since mid-December so I don't see a way that it could be relying on features from `grpcio` 1.18.0. However this might be a reason to pin `google-api-core` as well just in case.

BenRKarl:
Closing this PR in favor of: https://github.com/googleads/google-ads-python/pull/27, which makes the same changes in addition to some logic improvements.

Issue #30, opened 2019-01-31
querying video_quartile_25_rate causes google.protobuf.internal.well_known_types.Error: Fail to print FieldMask to Json string


apurvis:
This issue arose trying to upgrade to package 0.6/API 0.7. Trying to query `metrics.video_quartile_25_rate` gives **google.protobuf.internal.well_known_types.Error: Fail to print FieldMask to Json string: The character after a "_" must be a lowercase letter in path name metrics.video_quartile_25_rate.**

full stack trace:
```
File "/Users/oogabooga/repos/airflow-workflows/.venv/lib/python2.7/site-packages/google/api_core/page_iterator.py", line 199, in _items_iter
   for page in self._page_iter(increment=False):
 File "/Users/oogabooga/repos/airflow-workflows/.venv/lib/python2.7/site-packages/google/api_core/page_iterator.py", line 230, in _page_iter
   page = self._next_page()
 File "/Users/oogabooga/repos/airflow-workflows/.venv/lib/python2.7/site-packages/google/api_core/page_iterator.py", line 512, in _next_page
   response = self._method(self._request)
 File "/Users/oogabooga/repos/airflow-workflows/.venv/lib/python2.7/site-packages/google/api_core/gapic_v1/method.py", line 139, in __call__
   return wrapped_func(*args, **kwargs)
 File "/Users/oogabooga/repos/airflow-workflows/.venv/lib/python2.7/site-packages/google/api_core/retry.py", line 260, in retry_wrapped_func
   on_error=on_error,
 File "/Users/oogabooga/repos/airflow-workflows/.venv/lib/python2.7/site-packages/google/api_core/retry.py", line 177, in retry_target
   return target()
 File "/Users/oogabooga/repos/airflow-workflows/.venv/lib/python2.7/site-packages/google/api_core/timeout.py", line 206, in func_with_timeout
   return func(*args, **kwargs)
 File "/Users/oogabooga/repos/airflow-workflows/.venv/lib/python2.7/site-packages/google/api_core/grpc_helpers.py", line 59, in error_remapped_callable
   return callable_(*args, **kwargs)
 File "/Users/oogabooga/repos/airflow-workflows/.venv/lib/python2.7/site-packages/grpc/_interceptor.py", line 201, in __call__
   credentials=credentials)
 File "/Users/oogabooga/repos/airflow-workflows/.venv/lib/python2.7/site-packages/grpc/_interceptor.py", line 226, in _with_call
   return call.result(), call
[...]

PR #37, opened 2019-02-14
Temporarily patch FieldMask ToJsonString method

BenRKarl:
This is a temporary patch to fix [this Issue](https://github.com/googleads/google-ads-python/issues/30) until one of two things happens:

1. We refactor the logging library to not rely on serializing messages to JSON.
2. A permanent fix is introduced into the Python protobuf library.

### DIFF
--- a/google/ads/google_ads/util.py
+++ b/google/ads/google_ads/util.py
@@ -13,19 +13,19 @@
 # limitations under the License.
 
 
-# TODO - Remove implementation of MonkeyPatching ToJsonString once issue
+# TODO - Remove patching ToJsonString once issue
 # has been addressed in the protobuf library.
 def patch_to_json_method(obj):
-    """Monkey patches an object's ToJsonString method.
+    """Patches an object's ToJsonString method.
 
     Args:
-        obj: An arbitrary object.isinstance
+        obj: An arbitrary object.
     """
     obj.ToJsonString = _patched_to_json_string
 
 
 def _patched_to_json_string(self):
-    """Converts an objects paths array to a camel cased string."""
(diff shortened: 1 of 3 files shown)


google/ads/google_ads/util.py
```diff
@@ -0,0 +1,79 @@
...
+      raise Error('Fail to print FieldMask to Json string: Path name '
+                  '{0} must not contain uppercase letters.'.format(path_name))
+    if after_underscore:
+      if c.islower() or c.isdigit():
+        result.append(c.upper())
+        after_underscore = False
+      else:
+        raise Error('Fail to print FieldMask to Json string: The '
+                    'character after a "_" must be a lowercase letter '
+                    'in path name {0}.'.format(path_name))
```
msaniscalchi: Given that we're adding digits, probably safe to add them.

google/ads/google_ads/util.py
```diff
@@ -0,0 +1,79 @@
...
+    for path in paths:
+        camelcase_paths.append(_snake_case_to_camel_case(path))
+
+    return ','.join(camelcase_paths)
+
+
+def _snake_case_to_camel_case(path_name):
+  """Converts a path name from snake_case to camelCase.
+
+  This is a slightly modified version of the function defined in the protobuf
```
wihl: Would be useful to mention what modifications were made. Who knows when the original will be fixed.

BenRKarl:
I'm going to close this PR because I think a better change would be to update the logging logic so that it doesn't rely on converting proto messages to JSON, since the protos can be logged directly with basically the same level of detail.

PR #43, opened 2019-03-06
fix path in v1/__init__ file

BenRKarl:


ChangeLog
```diff
@@ -1,3 +1,6 @@
+* 1.0.1:
+- Fix bug in import path.
```
AnashOommen: You can point to the github bug here.

PR #46, opened 2019-03-07
Prepend "googleAdsField/" to `resource_name` if not found at start of string

MarTango:
Not sure if the code is automatically generated, but I find it a bit nicer to use `google_ads_field_service.get_google_ads_field("campaign")` instead of `.get_google_ads_field("googleAdsFields/campaign")`

Usage changes from:

```python
service = client.get_service("GoogleAdsFieldService")
service.get_google_ads_field("googleAdsFields/campaign")
```

to

```python
...
service.get_google_ads_field("campaign")
```

### DIFF
--- a/google/ads/google_ads/v1/services/google_ads_field_service_client.py
+++ b/google/ads/google_ads/v1/services/google_ads_field_service_client.py
@@ -211,6 +211,9 @@ class GoogleAdsFieldServiceClient(object):
                     client_info=self._client_info,
                 )
 
+        if resource_name[:15] != "googleAdsFields/":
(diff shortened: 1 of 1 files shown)


MarTango:
signed

>

BenRKarl:
Hi @MarTango thank you for taking the time to submit this change! Unfortunately this is a generated file so modifying it manually will not be maintainable (since the change will be overwritten anytime the file is generated). If you wouldn't mind creating a separate issue that describes the interface you'd like to have I can try to have it built into our generation logic to make it permanent.

PR #51, opened 2019-03-13
Add generate_keyword_ideas example

BenRKarl:


examples/planning/generate_keyword_ideas.py
```diff
@@ -0,0 +1,148 @@
...
+    locations = map_locations_to_string_values(client, location_ids)
+    language = map_language_to_string_value(client, language_id)
+
+    # Only one of these values will be passed to the KeywordPlanIdeaService
+    # depending on whether keywords, a page_url or both were given.
+    url_seed = None
+    keyword_seed = None
+    keyword_url_seed = None
+
+    if (not len(keywords) and not page_url):
```
msaniscalchi: *nitpick*
Generally good to use `not` in conditionals as infrequently as possible.

```not (len(keywords) or page_url):```

File: examples/planning/generate_keyword_ideas.py [hunk 2]
```diff
@@ -0,0 +1,148 @@
...
+        raise ValueError('At least one of keywords or page URL is required, '
+                         'but neither was specified.')
+    elif (not len(keywords) and page_url):
```
msaniscalchi: Is there a hypothetical world where these can all be set? If so, probably don't want to use `elif`.

examples/planning/generate_keyword_ideas.py
```diff
@@ -0,0 +1,148 @@
...
+
+    parser = argparse.ArgumentParser(
+        description='Generates keyword ideas from a list of seed keywords.')
+
+    # The following argument(s) should be provided to run the example.
+    parser.add_argument('-c', '--customer_id', type=six.text_type,
+                        required=True, help='The Google Ads customer ID.')
+    # For more information on determining location IDs, see:
+    parser.add_argument('-k', '--keywords', type=six.text_type, required=False,
+                        help='Comma-separated starter keywords')
```
msaniscalchi: Might as well be consistent. I think we frequently use "delimited" to describe this, and you do so below as well.

examples/planning/generate_keyword_ideas.py
BenRKarl: Done

examples/planning/generate_keyword_ideas.py
BenRKarl: Good point - I simplified this flow so it only uses if statements and added comments explaining each scenario.

examples/planning/generate_keyword_ideas.py
```diff
@@ -0,0 +1,162 @@
...
+
+from __future__ import absolute_import
+
+import argparse
+import six
+import sys
+from google.ads.google_ads.client import GoogleAdsClient
+from google.ads.google_ads.errors import GoogleAdsException
+
+_DEFAULT_LOCATION_IDS = '1023191'  # location ID for New York, NY
```
wihl: please add links to the documentation page where these literals are defined

examples/planning/generate_keyword_ideas.py
```diff
@@ -0,0 +1,162 @@
...
+
+    # Either keywords or a page_url are required to generate keyword ideas
+    # so this raises an error if neither are provided.
+    if (not (len(keywords) or page_url)):
+        raise ValueError('At least one of keywords or page URL is required, '
+                         'but neither was specified.')
+
+    # To generate keyword ideas with only a page_url and no keywords we need
+    # to initialize a UrlSeed object with the page_url as the "url" field.
+    if (not len(keywords) and page_url):
```
wihl: I recommend an extra set of parens to remove ambiguity, ie. if ((not len(keywords)) and page_url):

examples/planning/generate_keyword_ideas.py
```diff
@@ -0,0 +1,162 @@
...
+    # For more information on determining location IDs, see:
+    parser.add_argument('-k', '--keywords', type=six.text_type, required=False,
+                        help='Comma-delimited starter keywords')
+    # https://developers.google.com/adwords/api/docs/appendix/geotargeting.
+    parser.add_argument('-l', '--location_ids', type=six.text_type,
+                        required=False, help='Comma-delimited list of location '
+                                             'criteria IDs')
+    # A language criterion ID. For example, specify 1000 for English. For more
+    # information on determining this value, see the below link:
+    # https://developers.google.com/adwords/api/docs/appendix/codes-formats#languages.
```
wihl: You can move the comment containing the link to line 24 where it is more directly relevant

BenRKarl: Sounds good - I moved the comment up top but left the link for people who are passing these values in via the CLI.

examples/planning/generate_keyword_ideas.py
```diff
@@ -0,0 +1,162 @@
...
+    # to initialize a UrlSeed object with the page_url as the "url" field.
+    if (not len(keywords) and page_url):
+        url_seed = client.get_type('UrlSeed', version='v1')
+        url_seed.url.value = page_url
+
+
+    # To generate keyword ideas with only a list of keywords and no page_url
+    # we need to initialize a KeywordSeed object and set the "keywords" field
+    # to be a list of StringValue objects.
+    if (len(keywords) and not page_url):
```
msaniscalchi: *Nitpick*
You don't need to enclose your conditionals in parenthesis. Applies here and elsewhere.

BenRKarl: I took them out, looks much nicer now.

BenRKarl:
@wihl I think I've addressed everything here.

examples/planning/generate_keyword_ideas.py
```diff
@@ -0,0 +1,168 @@
...
+import argparse
+import six
+import sys
+from google.ads.google_ads.client import GoogleAdsClient
+from google.ads.google_ads.errors import GoogleAdsException
+
+# Location IDs are listed here: https://developers.google.com/adwords/api/docs/appendix/geotargeting
+# and they can also be retrieved using the GeoTargetConstantService as shown
+# here: https://developers.google.com/google-ads/api/docs/targeting/location-targeting
+_DEFAULT_LOCATION_IDS = '1023191'  # location ID for New York, NY
```
wihl: since the argument in the call is in plural form (_IDS), how would you specify more than one ID? ['1023191', '123']?

examples/planning/generate_keyword_ideas.py
```diff
@@ -0,0 +1,168 @@
...
+
+    # Only one of these values will be passed to the KeywordPlanIdeaService
+    # depending on whether keywords, a page_url or both were given.
+    url_seed = None
+    keyword_seed = None
+    keyword_url_seed = None
+
+    # Either keywords or a page_url are required to generate keyword ideas
+    # so this raises an error if neither are provided.
+    if not (len(keywords) or page_url):
```
wihl: a slightly better form is to remove the redundant len(), ie. if not (keywords or page_url):

see http://google.github.io/styleguide/pyguide.html#2144-decision

examples/planning/generate_keyword_ideas.py
```diff
@@ -0,0 +1,168 @@
...
+        raise ValueError('At least one of keywords or page URL is required, '
+                         'but neither was specified.')
+
+    # To generate keyword ideas with only a page_url and no keywords we need
+    # to initialize a UrlSeed object with the page_url as the "url" field.
+    if not len(keywords) and page_url:
```
wihl: again here, if not keywords and page_url:

examples/planning/generate_keyword_ideas.py
```diff
@@ -0,0 +1,168 @@
...
+        url_seed = client.get_type('UrlSeed', version='v1')
+        url_seed.url.value = page_url
+
+
+    # To generate keyword ideas with only a list of keywords and no page_url
+    # we need to initialize a KeywordSeed object and set the "keywords" field
+    # to be a list of StringValue objects.
+    if len(keywords) and not page_url:
```
wihl: if keywords and not page_url:

examples/planning/generate_keyword_ideas.py
```diff
@@ -0,0 +1,168 @@
...
+        keyword_seed = client.get_type('KeywordSeed', version='v1')
+        keyword_protos = map_keywords_to_string_values(client, keywords)
+        keyword_seed.keywords.extend(keyword_protos)
+
+    # To generate keyword ideas using both a list of keywords and a page_url we
+    # need to initialize a KeywordAndUrlSeed object, setting both the "url" and
+    # "keywords" fields.
+    if len(keywords) and page_url:
```
wihl: if keywords and page_url:

examples/planning/generate_keyword_ideas.py
```diff
@@ -0,0 +1,168 @@
...
+        keyword_ideas = keyword_plan_idea_service.generate_keyword_ideas(
+            customer_id, language, locations, keyword_plan_network,
+            url_seed=url_seed, keyword_seed=keyword_seed,
+            keyword_and_url_seed=keyword_url_seed)
+
+        for idea in keyword_ideas.results:
+            competition_value = keyword_competition_level_enum.Name(
+                idea.keyword_idea_metrics.competition)
+            print('Keyword idea text "%s" has %d average monthly searches and '
+                  '"%s" competition.\n' % (
```
wihl: s/%/.format

examples/planning/generate_keyword_ideas.py
```diff
@@ -0,0 +1,168 @@
...
+                      idea.text.value,
+                      idea.keyword_idea_metrics.avg_monthly_searches.value,
+                      competition_value))
+    except GoogleAdsException as ex:
+        print('Request with ID "%s" failed with status "%s" and includes the '
+              'following errors:' % (ex.request_id, ex.error.code().name))
```
wihl: s/%/.format

examples/planning/generate_keyword_ideas.py
```diff
@@ -0,0 +1,168 @@
...
+        for error in ex.failure.errors:
+            print('\tError with message "%s".' % error.message)
+            if error.location:
+                for field_path_element in error.location.field_path_elements:
+                    print('\t\tOn field: %s' % field_path_element.field_name)
```
wihl: s/%/.format

examples/planning/generate_keyword_ideas.py
BenRKarl: Since this is a default value for the command line input it would need to be a comma-separated list of strings.

examples/planning/generate_keyword_ideas.py
BenRKarl: Oh nice I didn't know an empty array is Falsy in this way.

BenRKarl:
@wihl everything is addressed save for your first comment. LMK if we should take action on it. For clarity I can change the default to a commad-delimited string for a stronger example, if you think it's necessary.

examples/planning/generate_keyword_ideas.py
wihl: it's a nit. If the variable name is plural, we should show more than one value. consider it for future reference.

wihl:
LGTM

PR #52, opened 2019-03-14
Add utility to format composite resource names

BenRKarl:
Instead of requiring composite ID strings to be concatenated by the user, this simply adds a utility that does it for them.

### DIFF
--- a/examples/basic_operations/pause_ad.py
+++ b/examples/basic_operations/pause_ad.py
@@ -19,8 +19,9 @@ import argparse
 import six
 import sys
 
-import google.ads.google_ads.client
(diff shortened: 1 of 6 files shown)


BenRKarl:
@msaniscalchi I agree it's a bit overkill, but my goal was to consolidate the logic so it isn't repeated across examples and has comments for documentation. Also I'd like to eventually build that class out to handle more of the construction logic for building resource names similar to how Java [does it](https://github.com/googleads/google-ads-java/blob/master/google-ads/src/main/java/com/google/ads/googleads/v1/utils/ResourceNames.java).

PR #59, opened 2019-03-21
added sitelink example

FBosler:


examples/advanced_operations/add_sitelink.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2018 Google LLC
```
BenRKarl: Can you update `2018` to `2019`

examples/advanced_operations/add_sitelink.py
```diff
@@ -0,0 +1,69 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Demonstrates how to create a sitelink extension."""
+
+from __future__ import absolute_import
+
+import argparse
+import six
+import sys
+import uuid
```
BenRKarl: This appears to be unused, can you remove it?

examples/advanced_operations/add_sitelink.py
```diff
@@ -0,0 +1,69 @@
...
+
+import google.ads.google_ads.client
```
BenRKarl: Can you update this to:

```
from google.ads.google_ads.client import GoogleAdsClient
```

And on line 59 you can just use:

```
google_ads_client = GoogleAdsClient.load_from_storage()
```

examples/advanced_operations/add_sitelink.py
```diff
@@ -0,0 +1,69 @@
...
+    extension_feed_item.sitelink_feed_item.line2.value = 'Line 2 Value'
+    final_url = extension_feed_item.sitelink_feed_item.final_urls.add()
+    final_url.value = 'www.example.com'
+
+    # Add extension
+    try:
+        feed_response = (
+            feed_service.mutate_extension_feed_items(customer_id, [extension_feed_item_operation])
+        )
+    except google.ads.google_ads.errors.GoogleAdsException as ex:
```
BenRKarl: Similar to my comment about `GoogleAdsClient` can you add this to the top of the file:

```
from google.ads.google_ads.errors import GoogleAdsException
```
Then this line can just be:

```
except GoogleAdsException as ex:
```

examples/advanced_operations/add_sitelink.py
FBosler: Hi Ben, thx for the feedback. I am not a 100% sure how you typically want to run those examples. Judging from the other examples though, it seems like they are supposed to be run from the command line.
This means we need this part to get the customer_id.
Which brings me to another question, I just realized is somewhat strange. Actually the client gets instanciated with a ```client_customer_id```. However, it seems like its not being used or being exposed for that matter.

examples/advanced_operations/add_sitelink.py
FBosler: much better, we should probably change that in the other examples as well.

examples/advanced_operations/add_sitelink.py
FBosler: also much better, we should adjust in the other examples as well. I can help, if you'd like me to.

examples/advanced_operations/add_sitelink.py
BenRKarl: Hi @FBosler thanks for making the updates! I'm definitely planning to make the style changes to the other examples at some point. If you have time and are interested please feel free to submit a PR but don't feel obligated.

The examples are indeed designed to be run from the command line. Since there can be variance in the CID and also inputs such as keywords, etc. it makes them much easier to work with. 

You're correct about the `client_customer_id` in that the client doesn't import it. In the client library for the AdWords API there was an option to include the CID in yaml, but that option doesn't exist in this library so you need to supply it explicitly. Is having the option to include it in yaml something you'd be interested in having? If so I can add a feature request and see if it's worth adding across the libraries.

BenRKarl: Actually @FBosler I created this issue to continue the discussion about CID so that I can merge this PR. https://github.com/googleads/google-ads-python/issues/new

PR #95, opened 2019-05-02
Remove JSON serialization of Proto messages

BenRKarl:
JSON Serialization of fields such as `video_quartile_25_rate` that contain a digit following an underscore will raise errors in Python.

This is expected behavior because it's not possible to serialize from snake_case to camelCase and back when a digit follows an underscore.

The fields in the API that contain this pattern are technically breaking this convention, but until a broader resolution is arrived at it's easier for now to simply not serialize messages to JSON for the purposes of logging because proto objects serialize to string very well on their own already.

This PR also introduces some general refactoring and cleanup.

### DIFF
--- a/google/ads/google_ads/client.py
+++ b/google/ads/google_ads/client.py
@@ -455,7 +455,7 @@ class LoggingInterceptor(grpc.UnaryUnaryClientInterceptor):
         return getattr(request, 'customer_id', None)
 
     def _parse_exception_to_str(self, exception):
-        """Parses response object to JSON.
+        """Parses response object to str for logging.
 
         Returns:
             A str representing a response or exception from the
--- a/google/ads/google_ads/client.py
+++ b/google/ads/google_ads/client.py
@@ -21,7 +21,6 @@ import json
 from collections import namedtuple
 from importlib import import_module
 
-import google.api_core.grpc_helpers
 import google.auth.transport.requests
 import google.oauth2.credentials
 import google.ads.google_ads.errors
@@ -37,6 +36,7 @@ _PROTO_TEMPLATE = '%s_pb2'
 _DEFAULT_TOKEN_URI = 'https://accounts.google.com/o/oauth2/token'
 _VALID_API_VERSIONS = ['v1']
 _DEFAULT_VERSION = _VALID_API_VERSIONS[0]
+_REQUEST_ID_KEY = 'request-id'
 
 class GoogleAdsClient(object):
     """Google Ads client used to configure settings and fetch services."""
@@ -223,7 +223,6 @@ class GoogleAdsClient(object):
 class ExceptionInterceptor(grpc.UnaryUnaryClientInterceptor):
     """An interceptor that wraps rpc exceptions."""
 
-    _REQUEST_ID_KEY = 'request-id'
     # Codes that are retried upon by google.api_core.
     _RETRY_STATUS_CODES = (
         grpc.StatusCode.INTERNAL, grpc.StatusCode.RESOURCE_EXHAUSTED)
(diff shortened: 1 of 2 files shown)


google/ads/google_ads/client.py
```diff
@@ -18,15 +18,14 @@
...
 from collections import namedtuple
 from importlib import import_module
 
-import google.api_core.grpc_helpers
-import google.auth.transport.requests
-import google.oauth2.credentials
-import google.ads.google_ads.errors
-from google.protobuf.json_format import MessageToJson
-import grpc
+from google.auth.transport.requests import Request
```
wihl: module_name and package_names should be lowercase https://google.github.io/styleguide/pyguide.html#316-naming (I should've caught MessageToJson previously)

google/ads/google_ads/client.py
```diff
@@ -454,35 +418,31 @@ def _get_customer_id(self, request):
...
         """
-        if exception:
-            # try to retrieve the .failure property of a GoogleAdsFailure.
-            failure = getattr(exception, 'failure', None)
-
-            if failure:
-                return _parse_message_to_json(failure)
-            else:
+        try:
+            # If the exception is from the Google Ads API the failure attribute
```
wihl: nit: "API, the failure" or "API *, then* the failure"

google/ads/google_ads/client.py
```diff
@@ -454,35 +418,31 @@ def _get_customer_id(self, request):
...
-                    return _parse_to_json(json.loads(debug_string))
-                except (AttributeError, ValueError):
-                    # if both attempts to retrieve serializable error data fail
-                    # then simply return an empty JSON string
-                    return '{}'
-        else:
-            return _parse_message_to_json(response.result())
+                # transport error with a .debug_error_string method and the
+                # returned JSON string will need to be formatted.
+                debug_string = exception.debug_error_string()
```
wihl: do you need a throwaway `debug_string` variable?

google/ads/google_ads/client.py
```diff
@@ -505,79 +465,75 @@ def _get_fault_message(self, exception):
...
         Args:
             method: The method of the request.
             customer_id: The customer ID associated with the request.
             metadata_json: A JSON str of initial_metadata.
             request_id: A unique ID for the request provided in the response.
-            request_json: A JSON str of the request message.
+            request: An instance of a request proto message.
             trailing_metadata_json: A JSON str of trailing_metadata.
-            response_json: A JSON str of the the response message.
+            response: A grpc.Call/grpc.Future instance.
```
wihl: can it be a grpc.Future or just a grpc.Call?

google/ads/google_ads/client.py
```diff
@@ -588,11 +544,9 @@ def intercept_unary_unary(self, continuation, client_call_details, request):
             A grpc.Call/grpc.Future instance representing a service response.
```
wihl: can it be grpc.Future?

tests/client_test.py
```diff
@@ -25,11 +25,11 @@
 import grpc
 from pyfakefs.fake_filesystem_unittest import TestCase as FileTestCase
 
-import google.ads.google_ads.client
+from google.ads.google_ads import client as Client
```
wihl: same as previous comment re: case for modules

google/ads/google_ads/client.py
BenRKarl: Unfortunately this is an external module that we can't control, so would you prefer we use `as`, like:

```python
from google.auth.transport.requests import Request as auth_request
```

google/ads/google_ads/client.py
BenRKarl: The response object inherits from both classes, so it's technically an instance of both.

google/ads/google_ads/client.py
BenRKarl: It's technically both, so I left it this way to be explicit.

tests/client_test.py
BenRKarl: Same as above, I can alias if you think it's easier to read.

google/ads/google_ads/client.py
BenRKarl: Nope, I changed it.

google/ads/google_ads/client.py
msaniscalchi: My opinion is that it is probably more confusing to modify the name of modules/classes/etc in an imported library. If they break the style, I don't think it should be up to us to fix it.

PR #112, opened 2019-05-12
Add add_gmail_ad example

BenRKarl:


examples/advanced_operations/add_gmail_ad.py
```diff
@@ -0,0 +1,133 @@
...
+    gmail_ad.teaser.description.value = 'Create your own adventure.'
+    gmail_ad.teaser.business_name.value = 'Interplanetary Ships'
+    gmail_ad.teaser.logo_image.value = image_resource_names[0]
+    gmail_ad.marketing_image.value = image_resource_names[1]
+    gmail_ad.marketing_image_headline.value = 'Travel'
+    gmail_ad.marketing_image_description.value = 'Take to the skies!'
+
+    final_url = ad_group_ad.ad.final_urls.add()
+    final_url.value = 'http://www.example.com'
+    ad_group_ad.ad.name.value = 'Gmail Ad #{}'.format(str(uuid4())[:15])
```
wihl: why [:15]? It cuts in the middle of time_hi_and_version. Should be either [:13] or [:18]. See https://en.wikipedia.org/wiki/Universally_unique_identifier#Format

BenRKarl: I took this from the Ruby example, I believe is just limits the length of the random string to 15 chars. This is only used to randomize the ad name so there aren't collisions. Since we use the entire uuid in other examples I'll take it out.

PR #117, opened 2019-05-21
Add link_manager_to_client example

BenRKarl:


examples/account_management/link_manager_to_client.py
```diff
@@ -0,0 +1,109 @@
...
+            customer_client_link
+        WHERE
+            customer_client_link.resource_name = "{}"
+    '''.format(resource_name)
+
+    ga_service = client.get_service('GoogleAdsService', version='v1')
+    response = ga_service.search(manager_customer_id, query=query)
+
+    for row in response.result:
+        manager_link_id = row.customer_client_link.manager_link_id
```
wihl: What happens when there are multiple managers? Won't this just get the last one? Is that intended behavior?

BenRKarl: That's correct, but since we're filtering by `resource_name` we know only a single row will be returned. Since we can only get an iterator back from `search` we still need to iterate on it in order to get results, even though we know it will only return a single row. I added some comments here so that it's clearer what's happening.

examples/account_management/link_manager_to_client.py
```diff
@@ -54,8 +54,9 @@ def main(client, customer_id, manager_customer_id):
               manager_customer_id, customer_id, resource_name))
 
     # Find the manager_link_id of the link we just created, so we can construct
-    # the resource name for the link from the client side.
-
+    # the resource name for the link from the client side. Note that since we
+    # are filtering by resource_name, a unique identifier, only one.
```
wihl: extra '.' at end of line

BenRKarl: Fixed

PR #132, opened 2019-06-18
Add utf-8 encoding to all generated files

BenRKarl:
This has been much-needed for a while and came along with the generation that updated the service transport signatures to allow us to increase default payload size.


BenRKarl:
@wihl I'll do that in a separate PR, these updates are all automated from the generator and I want to get them out separately so that they don't add complexity to future generated code PRs.

wihl:
LGTM

PR #148, opened 2019-07-08
Adjust example to add multiple expanded text ads

jsshandle:


examples/basic_operations/add_expanded_text_ads.py
```diff
@@ -63,8 +69,8 @@ def main(client, customer_id, ad_group_id):
                     print('\t\tOn field: %s' % field_path_element.field_name)
         sys.exit(1)
 
-    print('Created ad group ad %s.'
-          % ad_group_ad_response.results[0].resource_name)
+    for result in ad_group_ad_response.results:
+        print('Created ad group ad %s.' % result.resource_name)
```
BenRKarl: Could you use `.format` instead of the `%` operator? For example:

```python
print('Created ad group ad {}.'.format(result.resource_name))
```

jsshandle: Done!

PR #150, opened 2019-07-11
Update partial failure example so it can be parsed for documentation.

BenRKarl:


BenRKarl:
@fiboknacky thank you for catching that, I updated the example.

examples/error_handling/handle_partial_failure.py
```diff
@@ -27,9 +27,67 @@
...
+            print_results(client, ad_group_response)
+        else:
+            print('All operations completed successfully. No partial failure '
+                  'to show.')
+
+        # In the list of results, operations from the ad_group_operation list
+        # that failed will be represented as empty messages. This loop detects
+        # such empty messages and ignores them, while printing information about
+        # successful operations.
+        for message in ad_group_response.results:
```
fiboknacky: I believe this should be the part  of the printResults method too. Please see this method (https://github.com/googleads/google-ads-php/blob/master/examples/ErrorHandling/HandlePartialFailure.php#L201) -- It will print what failed and what succeeded. For those succeeded, it will print their resource names.

BenRKarl: Done.

examples/error_handling/handle_partial_failure.py
```diff
@@ -53,58 +92,40 @@ def main(client, customer_id, campaign_id):
...
-                # the above ad_group_operations list failed (by index number)
-                # as well as the error message and error code.
-                print('A partial failure at index %d occured.\n'
-                      'Error message: %s\nError code: %s' % (
-                          error.location.field_path_elements[0].index.value,
-                          error.message, error.error_code))
-        else:
-            print('All operations completed successfully. No partial failure '
-                  'to show.')
+def check_if_partial_failure_exists(response):
```
wihl: this function is awkwardly named. Maybe just `partial_failure_occurred` or just `partial_failure`.

This simplifies the later code `if partial_failure_occurred ...` instead of `if check_if_partial_failure_exists...`

BenRKarl: I do agree but I defer to @fiboknacky because the name of the method is consistent with the other examples. @fiboknacky is it OK if we rename this method to something like David describes or does it need to be named the way it is now?

fiboknacky: It's fine to change, but please make sure you modify accordingly in the DevSite guide.

fiboknacky: BTW, partial_failure_occurred or partial_failure_exists seems most natural to me.

BenRKarl: Done, went with `is_partial_failure_error_present` because it's nice for functions that return a boolean like this to start with `is`.

PR #155, opened 2019-07-16
Add set_ad_parameters example

BenRKarl:


examples/campaign_management/set_ad_parameters.py
```diff
@@ -0,0 +1,114 @@
...
+    # GoogleAdsClient will read the google-ads.yaml configuration file in the
+    # home directory if none is specified.
+    google_ads_client = GoogleAdsClient.load_from_storage()
+
+    # Initializes a command line argument parser.
+    parser = argparse.ArgumentParser(
+        description='Adds an ad group for specified customer and campaign id.')
+
+    # The following argument(s) should be provided to run the example.
+    parser.add_argument('-c', '--customer_id', type=six.text_type,
```
wihl: These arguments are longs in Java. Why not set `type=six.integer_types` so validation is done on the command line?

BenRKarl: `add_argument` expects the `type` param to callable, and `six.integer_types` is actually a tuple, (to support `long` and `int` in Python 2) so using it generates an error.

We could set the type to `int` (since there is no `long` type in Python 3). That would generate an error immediately if someone tried to use a non-numeric value for a customer ID. The problem then is that we'd need to convert the numeric value to an instance of `str` or `bytes` before passing them into the service methods, otherwise they'll throw an error.

Actually it looks like they do [exactly that](https://github.com/googleads/google-ads-java/blob/master/google-ads-examples/src/main/java/com/google/ads/googleads/examples/basicoperations/GetCampaigns.java#L92) in Java so I'm not sure why they require the CIDs to be Longs in the examples.

In any case I'd prefer to stick with strings for these since that's what the service methods require.

wihl: ok, let's stick with strings and not concern about consistency of type safeness right now.

PR #164, opened 2019-07-23
Fix formatting in the code block

artemrys:


artemrys:
I signed it!

README.rst
```diff
@@ -107,6 +107,8 @@ You can also retrieve it by exporting environment variables.
 
 * Required for OAuth2 Service Account Flow:
 
+.. code-block::bash
```
BenRKarl: Thank you for noticing this! I was just looking viewing the output of this change and noticed this code block and the one above it are still not displayed. I believe it's because there should be a space before `bash`, so:

```
.. code-block:: bash
```

Would you mind updating this line, and also the one on line 102?

artemrys: Yes, you are right.
Done.

PR #165, opened 2019-08-02
Shopping Listing Tree

akaashhazarika:
***Should Test with New ad group****


wihl:
Replaced by https://github.com/googleads/google-ads-python/pull/165

PR #170, opened 2019-08-22
Adds example demonstrating how to use ReachPlanService.

johnjameswhitman:


examples/planning/forecast_reach.py
```diff
@@ -0,0 +1,277 @@
...
+                product.plannable_product_code.value)
+        planned_product.budget_micros.value = product.budget_micros.value
+        product_mix.append(planned_product)
+
+    _request_reach_curve(
+            client, customer_id, product_mix, location_id, currency_code)
+
+
+def main(client, customer_id):
+    location_id = '2840'  # US
```
wihl: suggest you add link to reference list of location IDs, ie. https://developers.google.com/google-ads/api/reference/rpc/google.ads.googleads.v2.services#google.ads.googleads.v2.services.GeoTargetConstantService

examples/planning/forecast_reach.py
```diff
@@ -0,0 +1,277 @@
...
+    )
+
+    print('Currency, Cost, On-Target Reach, On-Target Imprs, Total Reach,'
+          ' Total Imprs, Products')
+    for point in response.reach_curve.reach_forecasts:
+        product_splits = []
+        for p in point.forecasted_product_allocations:
+            product_splits.append({
+                    p.plannable_product_code.value:
+                        p.budget_micros.value / 1.0e6})
```
wihl: better practice to make 1.0e6 a const, ie. at the top define
MICROS_TO_CURRENCY = 1.0e6

as different local currencies may have different conversions.

examples/planning/forecast_reach.py
```diff
@@ -0,0 +1,277 @@
...
+        print([
+                currency_code,
+                point.cost_micros.value / 1.0e6,
```
wihl: ditto prev comment

examples/planning/forecast_reach.py
```diff
@@ -0,0 +1,277 @@
...
+      location_id: Location ID returned by ListPlannableLocations.
+      currency_code: Three-character ISO 4217 currency code.
+      budget: Budget to allocate to the plan.
+    """
+    product_mix = []
+    product_splits = [('TRUEVIEW_IN_STREAM', 0.15), ('BUMPER', 0.85)]
+    for product, split in product_splits:
+        planned_product = client.get_type('PlannedProduct', version='v2')
+        planned_product.plannable_product_code.value = product
+        planned_product.budget_micros.value = math.trunc(budget * 1.0e6 * split)
```
wihl: ditto currency comment

examples/planning/forecast_reach.py
johnjameswhitman: Done -- Added this link and a reference to the `ListPlannableLocations` method in the `ReachPlanService` docs. Also updated the docstrings to match.

examples/planning/forecast_reach.py
johnjameswhitman: Since we use this to convert back / forth, just named it `ONE_MILLION`.

examples/planning/forecast_reach.py
johnjameswhitman: Done

examples/planning/forecast_reach.py
johnjameswhitman: Done

johnjameswhitman:
@wihl thanks for your review. PTAL at your earliest convenience.

-John

johnjameswhitman:
Thanks both!

PR #174, opened 2019-09-10
Add interface to inject external interceptors into service requests

BenRKarl:


google/ads/google_ads/client.py
```diff
@@ -203,6 +203,9 @@ def get_service(self, name, version=_DEFAULT_VERSION):
                 "CampaignService" to retrieve a CampaignServiceClient instance.
             version: a str indicating the version of the Google Ads API to be
                 used.
+            interceptors: an optional list of interceptors to include in
+                requests. NOTE: this parameter is not intended for external use
```
wihl: s/external use/non-Google use/

PR #184, opened 2019-11-05
Demo: Speed up v2.services.enums import 4.4x (save >50ms)

gpshead:
This speeds up the import of the `google.ads.google_ads.v2.services.enums` module by 4.4x (on my machine).  Dropping from 69ms to 15.5ms.

It retains the types being `enum.IntEnum` instances, lazily instantiating them upon first access.

Consider this PR a demonstration, I'm assuming it would not be merged as is.  Internally see b/143728538.

### DIFF
--- a/google/ads/google_ads/v2/services/enums.py
+++ b/google/ads/google_ads/v2/services/enums.py
@@ -17,10381 +17,11432 @@
 """Wrappers for protocol buffer enum types."""
 
 import enum
+import six
+
+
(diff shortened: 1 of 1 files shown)


BenRKarl:
Closing this because this was implemented in a separate PR [here](https://github.com/googleads/google-ads-python/blob/70fd625459cf21ba9d404f7a35d9c7cd6f053097/google/ads/google_ads/v3/services/enums.py). Thanks @gpshead for the suggestion!

PR #185, opened 2019-11-19
Add enum to reporting example

htsedebenham:
Retrieves the name of the match type to make the printed result easier to read


wihl:
@htsedebenham Thanks for your contribution! I've taken your enum addition and applied several other updates in PR https://github.com/googleads/google-ads-python/pull/224

PR #189, opened 2019-11-23
Fix get hotel ads performance report example

yunstanford:
Fix get hotel ads performance report example.


yunstanford:
@googlebot I signed it!

PR #205, opened 2020-01-28
Added add_prices.py feed extension example

wihl:
PHP equivalent: https://github.com/googleads/google-ads-php/pull/265

### DIFF
--- a/examples/extensions/add_prices.py
+++ b/examples/extensions/add_prices.py
@@ -24,19 +24,10 @@ import sys
 
 from google.ads.google_ads.client import GoogleAdsClient
 from google.ads.google_ads.errors import GoogleAdsException
-from google.ads.google_ads.v2.services.campaign_service_client import (
-    CampaignServiceClient
-)
 
 
 def main(client, customer_id, campaign_id):
-    """The main method that creates all necessary entities for the example.
-
-    Args:
-        client: an initialized GoogleAdsClient instance.
-        customer_id: a client customer ID str.
-        campaign_id: a campaign ID str.
-    """
+    """The main method that creates all necessary entities for the example."""
     # Create the price extension feed item
     price_feed_item = client.get_type('PriceFeedItem', version='v2')
     price_feed_item.type = (
@@ -49,34 +40,28 @@ def main(client, customer_id, campaign_id):
     price_feed_item.language_code.value = 'en'
 
     # To create a price extension, at least three price offerings are needed.
+    price_extension_price_unit_enum = (
+        client.get_type('PriceExtensionPriceUnitEnum'))
     price_feed_item.price_offerings.extend([
-        create_price_offer(client,
-                    'Scrubs',
-                    'Body Scrub, Salt Scrub',
-                    60000000,  # 60 USD
-                    'USD',
-                    client.get_type('PriceExtensionPriceUnitEnum').PER_HOUR,
-                    'http://www.example.com/scrubs',
-                    'http://m.example.com/scrubs'),
-        create_price_offer(client,
-                    'Hair Cuts',
-                    'Once a month',
-                    75000000,  # 75 USD
-                    'USD',
-                    client.get_type('PriceExtensionPriceUnitEnum').PER_MONTH,
-                    'http://www.example.com/haircuts',
-                    'http://m.example.com/haircuts'),
-        create_price_offer(client, 'Skin Care Package',
-                    'Four times a month',
-                    250000000,  # 250 USD
-                    'USD',
-                    client.get_type('PriceExtensionPriceUnitEnum').PER_MONTH,
-                    'http://www.example.com/skincarepackage')
+        _create_price_offer(client, 'Scrubs', 'Body Scrub, Salt Scrub',
+            60000000,  # 60 USD
+            'USD', price_extension_price_unit_enum.PER_HOUR,
+            'http://www.example.com/scrubs', 'http://m.example.com/scrubs'),
+        _create_price_offer(client, 'Hair Cuts', 'Once a month',
+            75000000,  # 75 USD
+            'USD', price_extension_price_unit_enum.PER_MONTH,
+            'http://www.example.com/haircuts',
+            'http://m.example.com/haircuts'),
+        _create_price_offer(client, 'Skin Care Package', 'Four times a month',
+            250000000,  # 250 USD
+            'USD', price_extension_price_unit_enum.PER_MONTH,
+            'http://www.example.com/skincarepackage')
     ])
 
     # Create a customer extension setting using the previously created
     # extension feed item. This associates the price extension to your
     # account.
+    campaign_service = client.get_service('CampaignService', version='v2')
     extension_feed_item_operation = (
         client.get_type('ExtensionFeedItemOperation', version='v2'))
     extension_feed_item = extension_feed_item_operation.create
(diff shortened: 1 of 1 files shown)


examples/extensions/add_prices.py
```diff
@@ -0,0 +1,171 @@
...
+campaigns, run basic_operations/get_campaigns.py
+"""
+
+
+import argparse
+import sys
+
+from google.ads.google_ads.client import GoogleAdsClient
+from google.ads.google_ads.errors import GoogleAdsException
+from google.ads.google_ads.v2.services.campaign_service_client import (
```
msaniscalchi: This shouldn't be necessary, you can just get the client like any other via the GoogleAdsClient.get_service method, no?

examples/extensions/add_prices.py
```diff
@@ -0,0 +1,171 @@
...
+    CampaignServiceClient
+)
+
+
+def main(client, customer_id, campaign_id):
+    """The main method that creates all necessary entities for the example.
```
msaniscalchi: I hate to be nitpicky, but we have historically excluded pydocs from simple methods in examples where their purpose is intuitively apparent. Most existing examples don't do this, but is there a push to change this going forward?

examples/extensions/add_prices.py
```diff
@@ -0,0 +1,171 @@
...
+                    'http://m.example.com/scrubs'),
+        create_price_offer(client,
+                    'Hair Cuts',
+                    'Once a month',
+                    75000000,  # 75 USD
+                    'USD',
+                    client.get_type('PriceExtensionPriceUnitEnum').PER_MONTH,
+                    'http://www.example.com/haircuts',
+                    'http://m.example.com/haircuts'),
+        create_price_offer(client, 'Skin Care Package',
```
msaniscalchi: *Nitpick*
I don't think there's anything wrong with providing multiple arguments on a line, but it does break consistency when compared to the previous lines.

examples/extensions/add_prices.py
```diff
@@ -0,0 +1,171 @@
...
+        client.get_type('ExtensionFeedItemOperation', version='v2'))
+    extension_feed_item = extension_feed_item_operation.create
+    extension_feed_item.extension_type = (
+        client.get_type('ExtensionTypeEnum').PRICE)
+    extension_feed_item.price_feed_item.CopyFrom(price_feed_item)
+    extension_feed_item.targeted_campaign.value = (
+        CampaignServiceClient.campaign_path(customer_id, campaign_id))
+    extension_feed_item.ad_schedules.extend([
+        create_ad_schedule_info(client,
+            client.get_type('DayOfWeekEnum', version='v2').SUNDAY,
```
msaniscalchi: Given that you use these several times, consider making a local variable for the enums and using that.

examples/extensions/add_prices.py
```diff
@@ -0,0 +1,171 @@
...
+            if error.location:
+                for field_path_element in error.location.field_path_elements:
+                    print('\t\tOn field: %s' % field_path_element.field_name)
+        sys.exit(1)
+
+    print('Created extension feed with resource name {}.'
+          .format(feed_response.results[0].resource_name))
+
+
+def create_price_offer(client, header, description, price_in_micros,
```
msaniscalchi: You probably don't intend to share these helper methods. I'd recommend representing top-level methods or fields in a module as protected in that case.

examples/extensions/add_prices.py
```diff
@@ -0,0 +1,171 @@
...
+    try:
+        feed_service = client.get_service('ExtensionFeedItemService', version='v2')
+        # Issues a mutate request to add the customer extension setting and
+        # print its information.
+        feed_response = (
+            feed_service.mutate_extension_feed_items(customer_id, 
+                [extension_feed_item_operation])
+        )
+    except GoogleAdsException as ex:
+        print('Request with ID "%s" failed with status "%s" and includes the '
```
msaniscalchi: *Nitpick*
Given that we've been adjusting new examples to use f-strings, we should continue that here.

examples/extensions/add_prices.py
wihl: good point. Fixed.

examples/extensions/add_prices.py
wihl: Given the number of arguments, I think it is cleaner this way.

examples/extensions/add_prices.py
wihl: fixed.

examples/extensions/add_prices.py
wihl: was trying to make pylint happy. removed.

examples/extensions/add_prices.py
wihl: Changed my mind after removing the extra get_types()

examples/extensions/add_prices.py
wihl: fixed

examples/extensions/add_prices.py
wihl: fixed

examples/extensions/add_prices.py
```diff
@@ -49,55 +40,43 @@ def main(client, customer_id, campaign_id):
...
-                    'http://www.example.com/haircuts',
-                    'http://m.example.com/haircuts'),
-        create_price_offer(client, 'Skin Care Package',
-                    'Four times a month',
-                    250000000,  # 250 USD
-                    'USD',
-                    client.get_type('PriceExtensionPriceUnitEnum').PER_MONTH,
-                    'http://www.example.com/skincarepackage')
+        _create_price_offer(client, 'Scrubs', 'Body Scrub, Salt Scrub',
+            60000000,  # 60 USD
```
msaniscalchi: There's a few ways to list a large number of arguments while following recommended Python style, but you're sort of mixing two approaches here. I'd recommend a look at [PEP8's description of indentation](https://www.python.org/dev/peps/pep-0008/#indentation) for guidance.

E.g. You could do:
```
        # If you really want to emphasize individual arguments.
        _create_price_offer(client,
                            'Scrubs',
                            'Body Scrub, Salt Scrub',
                            60000000,  # 60 USD
                            'USD',
                            price_extension_price_unit_enum.PER_HOUR,
                            'http://www.example.com/scrubs',
                            'http://m.example.com/scrubs'),
```

Or:

```
        _create_price_offer(client, 'Scrubs', 'Body Scrub, Salt Scrub',
                            60000000, 'USD',
                            price_extension_price_unit_enum.PER_HOUR,
                            'http://www.example.com/scrubs',
                            'http://m.example.com/scrubs'),
```

Or:

```
        _create_price_offer(
            client, 'Scrubs', 'Body Scrub, Salt Scrub', 60000000, 'USD',
            price_extension_price_unit_enum.PER_HOUR,
            'http://www.example.com/scrubs', 'http://m.example.com/scrubs'),
```

I generally prefer to go with the latter, but any of these would be fine, provided its consistent across the example.

wihl: I'm doing the latter, but I left the comment re: USD in to be consistent with the golden example.

msaniscalchi: It doesn't appear that you're doing the latter; you shouldn't have arguments on the first line in the case.

W.r.t. the golden CL's comment, maybe place it over the `_create_price_offer` call? I don't think precise placement is so critical that you'd need to do unconventional indentation to make it work.

PR #206, opened 2020-01-29
Add uploadimageasset

wihl:
Based on https://github.com/googleads/google-ads-dotnet/tree/master/examples/Misc

### DIFF
--- a/examples/misc/upload_image_assets.py
+++ b/examples/misc/upload_image_assets.py
@@ -46,7 +46,7 @@ def main(client, customer_id):
     # If you specify the name field, then both the asset name and the image
     # being uploaded should be unique, and should not match another ACTIVE
     # asset in this customer account.
-    # asset.name = 'Jupiter Trip #' + ExampleUtilities.GetRandomString(),
+    # asset.name = 'Jupiter Trip #' + uuid.uuid4()
 
     asset_service = client.get_service('AssetService', version='v2')
 
@@ -55,13 +55,13 @@ def main(client, customer_id):
             asset_service.mutate_assets(customer_id,
                                         [asset_operation])
         )
-        print(f'Uploaded file(s):')
+        print('Uploaded file(s):')
         for row in mutate_asset_response.results:
             print(f'\tResource name: {row.resource_name}')
 
     except GoogleAdsException as ex:
-        print('Request with ID "%s" failed with status "%s" and includes the '
-              'following errors:' % (ex.request_id, ex.error.code().name))
+        print(f'Request with ID "{ex.request_id}" failed with status '
+              f'"{ex.error.code().name}" and includes the following errors:')
         for error in ex.failure.errors:
             print('\tError with message "%s".' % error.message)
             if error.location:
--- a/examples/misc/upload_image_assets.py
+++ b/examples/misc/upload_image_assets.py
@@ -29,7 +29,7 @@ def main(client, customer_id):
     """Main method, to run this code example as a standalone application."""
 
     # Download image from URL
-    URL = "https://goo.gl/3b9Wfh"
+    URL = 'https://goo.gl/3b9Wfh'
     image_content = requests.get(URL).content
 
     asset_operation = client.get_type('AssetOperation', version='v2')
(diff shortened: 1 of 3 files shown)


examples/misc/upload_image_assets.py
```diff
@@ -0,0 +1,85 @@
...
+    asset.image_asset.mime_type = client.get_type('MimeTypeEnum').IMAGE_JPEG
+    # Use your favorite image library to determine dimensions
+    asset.image_asset.full_size.height_pixels.value = 315
+    asset.image_asset.full_size.width_pixels.value = 600
+    asset.image_asset.full_size.url.value = URL
+    # Optional: Provide a unique friendly name to identify your asset.
+    # If you specify the name field, then both the asset name and the image
+    # being uploaded should be unique, and should not match another ACTIVE
+    # asset in this customer account.
+    # asset.name = 'Jupiter Trip #' + ExampleUtilities.GetRandomString(),
```
msaniscalchi: *nit*
Is this an actual utility included with the samples? I don't see it in the examples directory. If not, I'd just go with something recognizable from the standard library like uuid. It would be easier to get this to run if the user uncommented this field.

examples/misc/upload_image_assets.py
```diff
@@ -0,0 +1,85 @@
...
+
+    asset_service = client.get_service('AssetService', version='v2')
+
+    try:
+        mutate_asset_response = (
+            asset_service.mutate_assets(customer_id,
+                                        [asset_operation])
+        )
+        print(f'Uploaded file(s):')
```
msaniscalchi: *nit*
This doesn't appear to need to be an f-string.

examples/misc/upload_image_assets.py
```diff
@@ -0,0 +1,85 @@
...
+        for row in mutate_asset_response.results:
+            print(f'\tResource name: {row.resource_name}')
+
+    except GoogleAdsException as ex:
+        print('Request with ID "%s" failed with status "%s" and includes the '
```
msaniscalchi: If we're going to use f-strings, let's be consistent with their usage.

examples/misc/upload_image_assets.py
```diff
@@ -0,0 +1,85 @@
...
+    image_content = requests.get(URL).content
+
+    asset_operation = client.get_type('AssetOperation', version='v2')
+    asset = asset_operation.create
+    asset.type = client.get_type('AssetTypeEnum', version='v2').IMAGE
+    asset.image_asset.data.value = image_content
+    asset.image_asset.file_size.value = len(image_content)
+    asset.image_asset.mime_type = client.get_type('MimeTypeEnum').IMAGE_JPEG
+    # Use your favorite image library to determine dimensions
+    asset.image_asset.full_size.height_pixels.value = 315
```
msaniscalchi: As an aside, I wonder if an interesting utility might be one that can identify the image type and size for a given file / URL. If all image assets across the API are represented in a similar way, it should be possible to even populate and return an image_asset that can be used for API calls such as this. Perhaps not in the scope of this change, however. :-)

examples/misc/upload_image_assets.py
wihl: fixed

examples/misc/upload_image_assets.py
wihl: fixed

examples/misc/upload_image_assets.py
wihl: fixed

examples/misc/upload_image_assets.py
wihl: There are various libraries (e.g. PIL) and pure Python solutions, but adding them would increase dependencies or distract from this sample so I left it hardcoded for now as an exercise for the viewer. :)

examples/misc/upload_image_assets.py
```diff
@@ -55,13 +55,13 @@ def main(client, customer_id):
...
         for row in mutate_asset_response.results:
             print(f'\tResource name: {row.resource_name}')
 
     except GoogleAdsException as ex:
-        print('Request with ID "%s" failed with status "%s" and includes the '
-              'following errors:' % (ex.request_id, ex.error.code().name))
+        print(f'Request with ID "{ex.request_id}" failed with status '
+              f'"{ex.error.code().name}" and includes the following errors:')
         for error in ex.failure.errors:
             print('\tError with message "%s".' % error.message)
```
msaniscalchi: Sorry for not being more clear, when I mentioned consistently using f-strings, I also meant that for every following line.

wihl: I think I got them all now at least in this file. There is a separate bug to track re: updating all the samples.

PR #208, opened 2020-01-29
Adding extensions/add_site_links.py example.

msaniscalchi:


examples/extensions/add_site_links.py
```diff
@@ -0,0 +1,219 @@
...
+              f'"{ex.error.code().name}" and includes the following errors:')
+        for error in ex.failure.errors:
+            print(f'\tError with message "{error.message}".')
+            if error.location:
+                for field_path_element in error.location.field_path_elements:
+                    print('\t\tOn field: {field_path_element.field_name}')
+        sys.exit(1)
+
+    print('Created CampaignExtensionSetting: '
+          f'"{response.results[0].resource_name}".')
```
wihl: nit: mixing strings and f-strings

examples/extensions/add_site_links.py
```diff
@@ -0,0 +1,219 @@
...
+                for field_path_element in error.location.field_path_elements:
+                    print(f'\t\tOn field: {field_path_element.field_name}')
+        sys.exit(1)
+
+    print('Created ExtensionFeedItems:')
+    for feed_item in feed_response.results:
+        print(f'\tResource name: {feed_item.resource_name}')
+
+    return [_to_string_value(client, resource_name) for resource_name in
+            [result.resource_name for result in feed_response.results]]
```
wihl: nested list comprehensions aren't very legible

examples/extensions/add_site_links.py
```diff
@@ -0,0 +1,219 @@
...
+        minute_of_hour_enum.ZERO, 21, minute_of_hour_enum.ZERO)
+    _populate_ad_schedule(
+        extension_feed_item4.ad_schedules.add(), day_of_week_enum.WEDNESDAY, 18,
+        minute_of_hour_enum.ZERO, 21, minute_of_hour_enum.ZERO)
+    _populate_ad_schedule(
+        extension_feed_item4.ad_schedules.add(), day_of_week_enum.THURSDAY, 18,
+        minute_of_hour_enum.ZERO, 21, minute_of_hour_enum.ZERO)
+    _populate_ad_schedule(
+        extension_feed_item4.ad_schedules.add(), day_of_week_enum.FRIDAY, 18,
+        minute_of_hour_enum.ZERO, 21, minute_of_hour_enum.ZERO)
```
wihl: maybe make this into a loop iterating on the day of week enum (since the rest is the same), so less repeated code?

examples/extensions/add_site_links.py
msaniscalchi: Yeah, admittedly, there isn't a whole lot of guidance on the style-correctness of this. Python automagically concatenates these, and one of them doesn't actually need any formatting, so making it an f-string doesn't *seem* necessary, imo. I'll change this if you feel strongly about it.

examples/extensions/add_site_links.py
msaniscalchi: Good idea.

examples/extensions/add_site_links.py
msaniscalchi: True, perhaps we shouldn't use those in an example.

PR #209, opened 2020-01-29
Adding the remarketing/add_remarketing_action.py example.

msaniscalchi:
This example has not been run yet w/ real data due to ongoing issue.

### DIFF
--- a/examples/remarketing/add_remarketing_action.py
+++ b/examples/remarketing/add_remarketing_action.py
@@ -21,7 +21,7 @@ associated tag snippets will be retrieved.
 
 import argparse
 import sys
-import uuid
+from uuid import uuid4
 
 import google.ads.google_ads.client
 
@@ -42,15 +42,12 @@ def _add_remarketing_action(client, customer_id):
     remarketing_action_service = client.get_service(
         'RemarketingActionService', version='v2')
 
-    # Create the operation.
     remarketing_action_operation = client.get_type(
         'RemarketingActionOperation', version='v2')
 
-    # Create remarketing action.
     remarketing_action = remarketing_action_operation.create
-    remarketing_action.name.value = f'Remarketing action #{uuid.uuid4()}'
+    remarketing_action.name.value = f'Remarketing action #{uuid4()}'
 
-    # Add the remarketing action.
     try:
         remarketing_action_response = (
             remarketing_action_service.mutate_remarketing_actions(
@@ -80,15 +77,11 @@ def _query_remarketing_action(client, customer_id, resource_name, page_size):
     google_ads_service_client = client.get_service(
         'GoogleAdsService', version='v2')
 
-    # Issues a search request by specifying page size.
     results = google_ads_service_client.search(
         customer_id, query=query, page_size=page_size)
 
     try:
-        for row in results:
-            # There is only one row because we limited the search using the
-            # resource name, which is unique.
-            return row.remarketing_action
+        return list(results)[0].remarketing_action
     except google.ads.google_ads.errors.GoogleAdsException as ex:
         print(f'Request with ID "{ex.request_id}" failed with status '
               f'"{ex.error.code().name}" and includes the following errors:')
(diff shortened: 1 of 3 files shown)


examples/remarketing/add_remarketing_action.py
```diff
@@ -0,0 +1,122 @@
...
+             'FROM remarketing_action '
+             'WHERE remarketing_action.resource_name = %s' % resource_name)
+
+    # Issues a search request by specifying page size.
+    response = google_ads_service_client.search(customer_id, query,
+                                                page_size=page_size)
+
+    # There is only one row because we limited the search using the resource
+    # name, which is unique.
+    return response.results[0]
```
wihl: when I run this example as is, I get `AttributeError: 'GRPCIterator' object has no attribute 'results'` on this line

msaniscalchi: ACK. It's unclear if the issue is entirely resolved. While there is no error when mutating, the following query is returning nothing for the associated `resource_name`.

msaniscalchi:
I've refactored this a bit and ran through it with real data.

examples/remarketing/add_remarketing_action.py
```diff
@@ -0,0 +1,143 @@
...
+        client, customer_id, remarketing_action_resource_name, page_size)
+
+    _print_remarketing_action_attributes(client, queried_remarketing_action)
+
+
+def _add_remarketing_action(client, customer_id):
+    remarketing_action_service = client.get_service(
+        'RemarketingActionService', version='v2')
+
+    # Create the operation.
```
wihl: I thought we are trying to get away from comments that state the obvious. Either this and the next comment should be removed, which is preferable, or another another trivial comment to the line above for consistency.

examples/remarketing/add_remarketing_action.py
```diff
@@ -0,0 +1,143 @@
...
+"""This example demonstrates usage of remarketing actions.
+
+A new remarketing action will be created for the specified customer, and its
+associated tag snippets will be retrieved.
+"""
+
+
+import argparse
+import sys
+import uuid
```
wihl: suggest you make this `from uuid import uuid4` and then reference just `uuid4()`

examples/remarketing/add_remarketing_action.py
```diff
@@ -0,0 +1,143 @@
...
+
+    google_ads_service_client = client.get_service(
+        'GoogleAdsService', version='v2')
+
+    # Issues a search request by specifying page size.
+    results = google_ads_service_client.search(
+        customer_id, query=query, page_size=page_size)
+
+    try:
+        for row in results:
```
wihl: why have a `for` loop at all if it is just one row? 

`return results[0].remarketing_action`

examples/remarketing/add_remarketing_action.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def _print_remarketing_action_attributes(client, remarketing_action):
+    tracking_code_type_enum = client.get_type(
+        'TrackingCodeTypeEnum', version='v2').TrackingCodeType
+    tracking_code_page_format_enum = client.get_type(
+        'TrackingCodePageFormatEnum', version='v2').TrackingCodePageFormat
+
+    print(f'Remarketing action has ID {remarketing_action.id.value} and name '
+          f'"{remarketing_action.name.value}". It has the following generated '
```
wihl: Suggest a line break, ie. `"f'"{remarketing_action.name.value}".\nIt..."`

examples/remarketing/add_remarketing_action.py
msaniscalchi: Because it's not a list or list-like object, it's a `GRPCIterator`. I wasn't sure iterating for the one remarketing action or using `list(results)[0].remarketing_action` was preferable, so I went with the option that didn't involve instantiating a new object. In hindsight though, maybe it'd be easier to read if I did.

PR #212, opened 2020-01-31
Validate Text Ad example

wihl:
Based on https://github.com/googleads/google-ads-dotnet/blob/master/examples/CampaignManagement/ValidateTextAd.cs


wihl:
The golden does not account for non-policy violation errors. I added additional error handling for this.

examples/campaign_management/validate_text_ad.py
```diff
@@ -0,0 +1,98 @@
...
+
+def main(client, customer_id, ad_group_id):
+    ad_group_ad_operation = client.get_type('AdGroupAdOperation', version='v2')
+    ad_group_ad = ad_group_ad_operation.create
+    ad_group_service = client.get_service('AdGroupService', version='v2')
+    ad_group_ad.ad_group.value = ad_group_service.ad_group_path(customer_id,
+        ad_group_id)
+    ad_group_ad.status = client.get_type('AdGroupAdStatusEnum',
+        version='v2').PAUSED
+    ad_group_ad.ad.expanded_text_ad.description.value = 'Luxury Cruise to Mars'
```
fiboknacky: Nit: Put an extra line above this one and add a comment "Create an expanded text ad object." or similar comment.

examples/campaign_management/validate_text_ad.py
```diff
@@ -0,0 +1,98 @@
...
+    ad_group_ad.ad.expanded_text_ad.headline_part1.value = (
+        'Visit the Red Planet in style.')
+    ad_group_ad.ad.expanded_text_ad.headline_part2.value = (
+        'Low-gravity fun for everyone!!')
+    final_url = ad_group_ad.ad.final_urls.add()
+    final_url.value = 'http://www.example.com/'
+
+    ad_group_ad_service = client.get_service('AdGroupAdService', version='v2')
+    # Attempt the mutate with validate_only=True
```
fiboknacky: Nit: comments end with a period.

examples/campaign_management/validate_text_ad.py
```diff
@@ -0,0 +1,98 @@
...
+            if (error.error_code.policy_finding_error ==
+                client.get_type('PolicyFindingErrorEnum',
+                                version='v2').POLICY_FINDING):
+                if error.details.policy_finding_details:
+                    count = 1
+                    details = (error.details.policy_finding_details
+                               .policy_topic_entries)
+                    for entry in details:
+                        print(f'{count}) Policy topic entry with topic '
+                              f'{entry.topic} and type {entry} '
```
fiboknacky: Please wrap entry topic and type with a quote too.

examples/campaign_management/validate_text_ad.py
wihl: done.

examples/campaign_management/validate_text_ad.py
wihl: fixed.

examples/campaign_management/validate_text_ad.py
wihl: fixed.

wihl: I did this but it looks weird because of the newlines in the entries:

```
1) Policy topic entry with topic "value: "SYMBOLS"
" and type "topic {
  value: "SYMBOLS"
}
type: PROHIBITED
evidences {
  text_list {
    texts {
      value: "!\n"
    }
  }
}
evidences {
  text_list {
    texts {
      value: "everyone!!"
    }
  }
}
" was found.
```

wihl: I fixed the print to be `type="{entry.type}" ` and it still looks weird. 

```
1) Policy topic entry with topic="value: "SYMBOLS"
" and type="2" was found.
```

Do you recommend: (a) mapping entry.type to the appropriate enum name or (b) just printing the whole `entry` without quotes. My vote is for (b) as the enums get mapped to names anyway. (I think the golden could use an update about this).

fiboknacky: I'm actually confused by this printing too.
Why do we want to print both entry and entry.topic?
entry.topic should be a part of entry?

So, why don't we just print {entry} (in the next line)?

WDYT?

wihl: Much better. Let's go with `print(f'{count}) Policy topic entry: \n{entry}\n')`:

```
1) Policy topic entry: 
topic {
  value: "SYMBOLS"
}
type: PROHIBITED
evidences {
  text_list {
    texts {
      value: "!\n"
    }
  }
}
evidences {
  text_list {
    texts {
      value: "everyone!!"
    }
  }
}

```

fiboknacky: Great

examples/campaign_management/validate_text_ad.py
```diff
@@ -0,0 +1,98 @@
...
+                                version='v2').POLICY_FINDING):
+                if error.details.policy_finding_details:
+                    count = 1
+                    details = (error.details.policy_finding_details
+                               .policy_topic_entries)
+                    for entry in details:
+                        print(f'{count}) Policy topic entry: \n{entry}\n')
+                    count += 1
+            else:
+                print(f'\tNon-policy finding error with message '
```
BenRKarl: This doesn't need to be an f-string since there aren't any replacements in it.

wihl: noted.

PR #215, opened 2020-02-04
add_app_campaign golden example

wihl:
Google Ads API golden version of https://github.com/googleads/googleads-python-lib/blob/master/examples/adwords/v201809/advanced_operations/add_universal_app_campaign.py

### DIFF
--- a/examples/advanced_operations/add_app_campaign.py
+++ b/examples/advanced_operations/add_app_campaign.py
@@ -12,7 +12,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""This example adds a Universal App campaign.
+"""This example adds an App campaign.
+
+See https://developers.google.com/google-ads/api/docs/app-campaigns/overview
 
 To get campaigns, run basic_operations/get_campaigns.py.
 To upload image assets for this campaign, run misc/upload_image.py.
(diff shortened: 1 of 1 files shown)


wihl:
Based on https://github.com/googleads/googleads-python-lib/blob/master/examples/adwords/v201809/advanced_operations/add_universal_app_campaign.py

examples/advanced_operations/add_app_campaign.py
```diff
@@ -0,0 +1,305 @@
...
+        .get_type('BiddingStrategyTypeEnum', version='v2').TARGET_CPA)
+    # Set the target CPA to $1 / app install.
+    #
+    # campaign_bidding_strategy is a 'oneof' message so setting target_cpa
+    # is mutually exclusive with other bidding strategies such as
+    # manual_cpc, commission, maximize_conversions, etc.
+    # See https://developers.google.com/google-ads/api/reference/rpc
+    # under current version / resources / Campaign
+    campaign.target_cpa.target_cpa_micros.value = 1000000
+    # Set the App Campaign Settings. Other settings are on a per-ad basis.
```
mcloonan: I think the sentence "Other settings are on a per-ad basis." requires the context that this is different from how it used to be. Either be more explicit about what those "other settings" are or remove.

examples/advanced_operations/add_app_campaign.py
```diff
@@ -0,0 +1,305 @@
...
+    campaign.app_campaign_setting.app_id.value = 'com.labpixies.colordrips'
```
mcloonan: Kind of weird that we have an actual existing app here. It could be taken as an endorsement. Can we change this to something generic? (Would that cause the API to reject the input?)

wihl: I took this from the AdWords API example

mcloonan: I understand that but I don't think it's a good idea to propagate it forward.

examples/advanced_operations/add_app_campaign.py
```diff
@@ -0,0 +1,306 @@
...
+    campaign_service = client.get_service('CampaignService', version='v2')
+    campaign_operation = client.get_type('CampaignOperation', version='v2')
+    campaign = campaign_operation.create
+    campaign.name.value = f'Interplanetary Cruise App #{uuid4()}'
+    campaign.campaign_budget.value = budget_resource_name
+    # Recommendation: Set the campaign to PAUSED when creating it to 
+    # prevent the ads from immediately serving. Set to ENABLED once you've
+    # added targeting and the ads are ready to serve.
+    campaign.status = (client.get_type('CampaignStatusEnum', version='v2')
+        .PAUSED)
```
BenRKarl: It's probably preferable to omit the outer parentheses and do this:

```
client.get_type(
    'CampaignStatusEnum', version='v2').PAUSED
```

examples/advanced_operations/add_app_campaign.py
```diff
@@ -0,0 +1,306 @@
...
+    # eligible to appear on multiple channels.
+    campaign.advertising_channel_type = client.get_type(
+        'AdvertisingChannelTypeEnum', version='v2').MULTI_CHANNEL
+    campaign.advertising_channel_sub_type = client.get_type(
+        'AdvertisingChannelSubTypeEnum', version='v2').APP_CAMPAIGN
+    # Define the bidding strategy during campaign creation. An App campaign
+    # cannot use a portfolio bidding strategy.
+    # App campaigns only support the TARGET_CPA bidding strategy.
+    campaign.bidding_strategy_type = (client
+        .get_type('BiddingStrategyTypeEnum', version='v2').TARGET_CPA)
```
BenRKarl: Same I think the better indentation is:

```
client.get_type(
    'BiddingStrategyTypeEnum', version='v2').TARGET_CPA
```

examples/advanced_operations/add_app_campaign.py
```diff
@@ -0,0 +1,306 @@
...
+    # is mutually exclusive with other bidding strategies such as
+    # manual_cpc, commission, maximize_conversions, etc.
+    # See https://developers.google.com/google-ads/api/reference/rpc
+    # under current version / resources / Campaign
+    campaign.target_cpa.target_cpa_micros.value = 1000000
+    # Set the App Campaign Settings.
+    campaign.app_campaign_setting.app_id.value = (
+        'com.google.android.apps.adwords')
+    campaign.app_campaign_setting.app_store = (client.get_type(
+        'AppCampaignAppStoreEnum', version='v2').GOOGLE_APP_STORE)
```
BenRKarl: Don't need outer parens here.

examples/advanced_operations/add_app_campaign.py
```diff
@@ -0,0 +1,306 @@
...
+    # Optional fields
+    campaign.start_date.value = (datetime.now() +
+        timedelta(1)).strftime('%Y%m%d')
+    campaign.end_date.value = (datetime.now() +
+        timedelta(365)).strftime('%Y%m%d')
+    # Optimize this campaign for getting new users for your app.
+    campaign.app_campaign_setting.bidding_strategy_goal_type = (client
+        .get_type('AppCampaignBiddingStrategyGoalTypeEnum',
+                  version='v2').OPTIMIZE_INSTALLS_TARGET_INSTALL_COST)
```
BenRKarl: I don't think you need the outer parens here.

examples/advanced_operations/add_app_campaign.py
```diff
@@ -0,0 +1,306 @@
...
+    campaign_criterion_operation = (client.get_type(
+        'CampaignCriterionOperation', version='v2'))
+    geo_target_constant_service = (client.get_service(
+        'GeoTargetConstantService', version='v2'))
+    language_constant_service = (client.get_service(
+        'LanguageConstantService', version='v2'))
+    location_type = (client.get_type('CriterionTypeEnum', version='v2')
+        .LOCATION)
+    language_type = (client.get_type('CriterionTypeEnum', version='v2')
+        .LANGUAGE)
```
BenRKarl: You shouldn't need outer parens in lines 186 - 194

examples/advanced_operations/add_app_campaign.py
wihl: it is necessary due to the `.get_type` method call on the subsequent line

BenRKarl: Will you go over the character limit if you use:

 ```
client.get_type(
    'AppCampaign ...
```

wihl: yes

PR #216, opened 2020-02-05
Use get_message helper when exporting dependent types for v2

BenRKarl:
Retrieving class names using `getmembers` on pb2 modules returns _all_ classes, including some which are not protobuf message classes, (namely `google.longrunning.operations_pb2.OperationsStub`). Since we use types.py to consolidate protobuf message classes for easy access by the `client.get_type` method we want to use `get_messages` to limit the exposed classes to just protobuf message classes.

### DIFF
--- a/google/ads/google_ads/v2/types.py
+++ b/google/ads/google_ads/v2/types.py
@@ -1635,7 +1635,7 @@ DEPENDENT_MODULE_LIST = [
 
 def _get_class_from_module(module_name):
     module = importlib.import_module(module_name)
-    for class_name, _ in get_messages(module).items(): # from inspect module
+    for class_name in get_messages(module).keys(): # from inspect module
         yield class_name
 
 def _populate_dependent_classes(module_list = DEPENDENT_MODULE_LIST):
--- a/google/ads/google_ads/v2/types.py
+++ b/google/ads/google_ads/v2/types.py
@@ -16,9 +16,7 @@
 
 
 import importlib
-import re
 import sys
-from inspect import getmembers, isclass
 from itertools import chain
 
 from google.api_core.protobuf_helpers import get_messages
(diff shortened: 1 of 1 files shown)


BenRKarl:
Resolves: https://github.com/googleads/google-ads-python/issues/217

google/ads/google_ads/v2/types.py
```diff
@@ -1637,7 +1635,7 @@
 
 def _get_class_from_module(module_name):
     module = importlib.import_module(module_name)
-    for class_name, _ in getmembers(module, isclass): # from inspect module
+    for class_name, _ in get_messages(module).items(): # from inspect module
```
wihl: The key line in get_messages is: 
```
if inspect.isclass(candidate) and issubclass(candidate, message.Message):
```
We can avoid another import if we just use that. 
Ref: https://github.com/googleapis/google-cloud-python/blob/dfffcc7c0a9208e00065378c6a59167a817c51f7/api_core/google/api_core/protobuf_helpers.py#L116

BenRKarl: I'd prefer to use the helper method as-is since we already use it elsewhere in the file, and in the future it's possible the rules for how these classes are filtered could change.

wihl: Thinking about it a bit more, I think we should go with my proposed solution, otherwise each module may be imported twice, leading to potential performance parity issues.

BenRKarl: How can the modules be imported twice? I'm not seeing how that would happen.

wihl: ok, let's proceed with `get_messages()`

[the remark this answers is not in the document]
BenRKarl: Good observation, no reason so I changed it to `keys`.

PR #223, opened 2020-02-15
get_responsive_search_ads golden (also includes missing add_responsive_search_ad example)

wihl:
Google Ads API golden version of https://github.com/googleads/googleads-python-lib/blob/master/examples/adwords/v201809/basic_operations/get_responsive_search_ads.py

Also added add_responsive_search_ad.py because it was missing from Python

### DIFF
--- a/examples/basic_operations/add_responsive_search_ad.py
+++ b/examples/basic_operations/add_responsive_search_ad.py
@@ -30,15 +30,15 @@ def main(client, customer_id, ad_group_id):
     ad_group_ad_service = client.get_service('AdGroupAdService', version='v2')
     ad_group_service = client.get_service('AdGroupService', version='v2')
 
-    pinned_headline = client.get_type('AdTextAsset', version='v2')
-    pinned_headline.text.value = f'Cruise to Mars #{str(uuid4())[:8]}'
     # Set a pinning to always choose this asset for HEADLINE_1. Pinning is
     # optional; if no pinning is set, then headlines and descriptions will be
     # rotated and the ones that perform best will be used more often.
-    pinned_headline.pinned_field = client.get_type(
-       'ServedAssetFieldTypeEnum', version='v2').HEADLINE_1
+    pinned_headline = _create_ad_text_asset(client,
+        f'Cruise to Mars #{str(uuid4())[:8]}',
+        client.get_type(
+            'ServedAssetFieldTypeEnum', version='v2').HEADLINE_1)
 
-    # Create the ad group ad
+    # Create the ad group ad.
     ad_group_ad_operation = client.get_type('AdGroupAdOperation', version='v2')
     ad_group_ad = ad_group_ad_operation.create
     ad_group_ad.status = client.get_type(
@@ -46,7 +46,7 @@ def main(client, customer_id, ad_group_id):
     ad_group_ad.ad_group.value = ad_group_service.ad_group_path(
         customer_id, ad_group_id)
 
-    # Set responsive search ad info
+    # Set responsive search ad info.
     final_url = ad_group_ad.ad.final_urls.add()
     final_url.value = 'http://www.example.com'
     ad_group_ad.ad.responsive_search_ad.headlines.extend([
@@ -59,6 +59,7 @@ def main(client, customer_id, ad_group_id):
     ad_group_ad.ad.responsive_search_ad.path1.value = 'all-inclusive'
     ad_group_ad.ad.responsive_search_ad.path2.value = 'deals'
 
+    # Send a request to the server to add a responsive search ad.
     try:
         ad_group_ad_response = ad_group_ad_service.mutate_ad_group_ads(
             customer_id, [ad_group_ad_operation])
@@ -73,12 +74,16 @@ def main(client, customer_id, ad_group_id):
         sys.exit(1)
 
     for result in ad_group_ad_response.results:
-        print(f'Created ad group ad {result.resource_name}')
+        print(f'Created responsive search ad with resource name '
+              f'"{result.resource_name}".')
 
 
-def _create_ad_text_asset(client, text):
+def _create_ad_text_asset(client, text, pinned_field=None):
+    """Create an AdTextAsset."""
     ad_text_asset = client.get_type('AdTextAsset', version='v2')
     ad_text_asset.text.value = text
+    if pinned_field:
+        ad_text_asset.pinned_field = pinned_field
     return ad_text_asset
 
 
--- a/examples/basic_operations/add_responsive_search_ad.py
+++ b/examples/basic_operations/add_responsive_search_ad.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""This example adds a responsive search ad to a given ad group.
+
+To get ad groups, run get_ad_groups.py.
+"""
+
+
+import argparse
+import sys
+from uuid import uuid4
+
+from google.ads.google_ads.client import GoogleAdsClient
+from google.ads.google_ads.errors import GoogleAdsException
+
+
+def main(client, customer_id, ad_group_id):
+    ad_group_ad_service = client.get_service('AdGroupAdService', version='v2')
+    ad_group_service = client.get_service('AdGroupService', version='v2')
+
+    pinned_headline = client.get_type('AdTextAsset', version='v2')
+    pinned_headline.text.value = f'Cruise to Mars #{str(uuid4())[:8]}'
+    # Set a pinning to always choose this asset for HEADLINE_1. Pinning is
+    # optional; if no pinning is set, then headlines and descriptions will be
+    # rotated and the ones that perform best will be used more often.
+    # pinned_headline.pinned_field = client.get_type(
+    #    'ServedAssetFieldTypeEnum', version='v2').HEADLINE_1
+
+    # Create the ad group ad
+    ad_group_ad_operation = client.get_type('AdGroupAdOperation', version='v2')
+    ad_group_ad = ad_group_ad_operation.create
+    ad_group_ad.status = client.get_type(
+        'AdGroupAdStatusEnum', version='v2').PAUSED
+    ad_group_ad.ad_group.value = ad_group_service.ad_group_path(
+        customer_id, ad_group_id)
+
+    # Set responsive search ad info
+    final_url = ad_group_ad.ad.final_urls.add()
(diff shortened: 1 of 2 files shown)


wihl:
Note that this example still uses paging because streaming is not available for Python.

fiboknacky:
The add_responsive_search_ad example is not in the plan at first, because the old one uses legacy responsive search ad.
I see no harm to add the example here too though. But I wonder if you could also update the PR name? It's confusing to me when I first jump to the files but see the add instead of get example.

examples/basic_operations/add_responsive_search_ad.py
```diff
@@ -0,0 +1,100 @@
...
+        pinned_headline,
+        _create_ad_text_asset(client, 'Best Space Cruise Line'),
+        _create_ad_text_asset(client, 'Experience the Stars')])
+    ad_group_ad.ad.responsive_search_ad.descriptions.extend([
+        _create_ad_text_asset(client, 'Buy your tickets now'),
+        _create_ad_text_asset(client, 'Visit the Red Planet')])
+    ad_group_ad.ad.responsive_search_ad.path1.value = 'all-inclusive'
+    ad_group_ad.ad.responsive_search_ad.path2.value = 'deals'
+
+    try:
```
fiboknacky: Nit: Let's add comments similar to the above like "Send a request to the server to add a responsive search ad"

examples/basic_operations/add_responsive_search_ad.py
```diff
@@ -0,0 +1,100 @@
...
+
+    pinned_headline = client.get_type('AdTextAsset', version='v2')
+    pinned_headline.text.value = f'Cruise to Mars #{str(uuid4())[:8]}'
+    # Set a pinning to always choose this asset for HEADLINE_1. Pinning is
+    # optional; if no pinning is set, then headlines and descriptions will be
+    # rotated and the ones that perform best will be used more often.
+    pinned_headline.pinned_field = client.get_type(
+       'ServedAssetFieldTypeEnum', version='v2').HEADLINE_1
+
+    # Create the ad group ad
```
fiboknacky: Comments should end with periods. Same for below.

examples/basic_operations/add_responsive_search_ad.py
```diff
@@ -0,0 +1,100 @@
...
+    ad_group_ad_service = client.get_service('AdGroupAdService', version='v2')
+    ad_group_service = client.get_service('AdGroupService', version='v2')
+
+    pinned_headline = client.get_type('AdTextAsset', version='v2')
+    pinned_headline.text.value = f'Cruise to Mars #{str(uuid4())[:8]}'
+    # Set a pinning to always choose this asset for HEADLINE_1. Pinning is
+    # optional; if no pinning is set, then headlines and descriptions will be
+    # rotated and the ones that perform best will be used more often.
+    pinned_headline.pinned_field = client.get_type(
+       'ServedAssetFieldTypeEnum', version='v2').HEADLINE_1
```
fiboknacky: Can't you make this be created by _create_ad_text_asset?
You probably can add one more param to choose whether to make the created text asset pinned or not.
This will head readability too because it will be closer to where it's used.

examples/basic_operations/add_responsive_search_ad.py
```diff
@@ -0,0 +1,100 @@
...
+              f'"{ex.error.code().name}" and includes the following errors:')
+        for error in ex.failure.errors:
+            print(f'\tError with message "{error.message}".')
+            if error.location:
+                for field_path_element in error.location.field_path_elements:
+                    print(f'\t\tOn field: {field_path_element.field_name}')
+        sys.exit(1)
+
+    for result in ad_group_ad_response.results:
+        print(f'Created ad group ad {result.resource_name}')
```
fiboknacky: Nit: Create a responsive search ad with resource name "{result.resource_name}".

examples/basic_operations/get_responsive_search_ads.py
```diff
@@ -0,0 +1,108 @@
...
+
+def main(client, customer_id, page_size, ad_group_id=None):
+    ga_service = client.get_service('GoogleAdsService', version='v2')
+
+    query = ('SELECT ad_group.id, ad_group_ad.ad.id, '
+             'ad_group_ad.ad.responsive_search_ad.headlines, '
+             'ad_group_ad.ad.responsive_search_ad.descriptions, '
+             'ad_group_ad.status FROM ad_group_ad '
+             'WHERE ad_group_ad.ad.type = RESPONSIVE_SEARCH_AD '
+             'AND ad_group_ad.status != ''REMOVED''')
```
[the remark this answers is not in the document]
wihl: Switched to [multiline string](http://google.github.io/styleguide/pyguide.html#310-strings) so it is easier for other client libs to cut'n'paste.

examples/basic_operations/get_responsive_search_ads.py
```diff
@@ -0,0 +1,108 @@
...
+                print('Headlines:\n{}\n'
+                      'Descriptions:\n{}\n'.format(
+                          '\n'.join(_ad_text_assets_to_strs(
+                              client, responsive_search_ad_info.headlines)),
+                          '\n'.join(_ad_text_assets_to_strs(
+                              client, responsive_search_ad_info.descriptions))))
+            else:
+                print('\tResponsive search ad info was not found.')
+
+        if not one_found:
```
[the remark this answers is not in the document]
wihl: No, `results` is a [GRPC iterator](https://googleapis.dev/python/google-api-core/latest/page_iterator.html) which is not a falsey. Data is returned lazily so the only way to know if there are >0 results is to start iterating. I'd rather leave in `one_found` because I think it is important to illustrate and show a message when there are zero results rather than ending silently.

examples/basic_operations/add_responsive_search_ad.py
wihl: fixed.

examples/basic_operations/add_responsive_search_ad.py
wihl: fixed.

examples/basic_operations/add_responsive_search_ad.py
wihl: fixed.

examples/basic_operations/add_responsive_search_ad.py
wihl: fixed.

examples/basic_operations/add_responsive_search_ad.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def _create_ad_text_asset(client, text):
```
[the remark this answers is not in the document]
wihl: I added one though in general we are trying to avoid overly simplistic comments.

examples/basic_operations/get_responsive_search_ads.py
```diff
@@ -0,0 +1,108 @@
...
+        for error in ex.failure.errors:
+            print(f'\tError with message "{error.message}".')
+            if error.location:
+                for field_path_element in error.location.field_path_elements:
+                    print(f'\t\tOn field: {field_path_element.field_name}')
+        sys.exit(1)
+
+
+def _ad_text_assets_to_strs(client, assets):
+    sa_field_type_enum = client.get_type(
```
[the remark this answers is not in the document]
wihl: added.

examples/basic_operations/get_responsive_search_ads.py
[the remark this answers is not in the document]
fiboknacky: Got it. Thanks.

examples/basic_operations/add_responsive_search_ad.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def main(client, customer_id, ad_group_id):
+    ad_group_ad_service = client.get_service('AdGroupAdService', version='v2')
+    ad_group_service = client.get_service('AdGroupService', version='v2')
+
+    # Set a pinning to always choose this asset for HEADLINE_1. Pinning is
+    # optional; if no pinning is set, then headlines and descriptions will be
+    # rotated and the ones that perform best will be used more often.
+    pinned_headline = _create_ad_text_asset(client,
```
fiboknacky: Can you move this line to after line 51?
It's not used anywhere around here and I don't see why we need to assign this variable so early in this function.

wihl: fixed.

examples/basic_operations/get_responsive_search_ads.py
```diff
@@ -0,0 +1,110 @@
...
+from google.ads.google_ads.errors import GoogleAdsException
+
+
+_DEFAULT_PAGE_SIZE = 1000
+
+
+def main(client, customer_id, page_size, ad_group_id=None):
+    ga_service = client.get_service('GoogleAdsService', version='v2')
+
+    query = '''
```
msaniscalchi: **Nitpick**
I don't feel strongly about it, but I generally recommend against the usage of triple-quote strings for most cases because it includes all the whitespace used for proper indentation in Python needlessly as part of the string. An equivalent would be something like:

```
    query = ('SELECT ad_group.id, ad_group_ad.ad.id, '
             'ad_group_ad.ad.responsive_search_ad.headlines, '
             'ad_group_ad.ad.responsive_search_ad.descriptions, '
             'ad_group_ad.status FROM ad_group_ad '
             'WHERE ad_group_ad.ad.type = RESPONSIVE_SEARCH_AD '
             'AND ad_group_ad.status != "REMOVED"')
```

examples/basic_operations/add_responsive_search_ad.py
```diff
@@ -0,0 +1,106 @@
...
+        customer_id, ad_group_id)
+
+    # Set responsive search ad info.
+    final_url = ad_group_ad.ad.final_urls.add()
+    final_url.value = 'http://www.example.com'
+
+    # Set a pinning to always choose this asset for HEADLINE_1. Pinning is
+    # optional; if no pinning is set, then headlines and descriptions will be
+    # rotated and the ones that perform best will be used more often.
+    pinned_headline = _create_ad_text_asset(client,
```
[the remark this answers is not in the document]
wihl: fixed

examples/basic_operations/get_responsive_search_ads.py
wihl: It was like that previously but I updated per the prior discussion. I'm going to leave it as is for now.

j-walker23: Random user of this python lib here. 
After months of hating writing queries in the original example way, and then having triple quotes + f-strings dawn on me.
From everyday dev writing happiness, triple quotes FTW! Thank you

examples/basic_operations/get_responsive_search_ads.py
```diff
@@ -0,0 +1,110 @@
...
+        one_found = False
+
+        for row in results:
+            one_found = True
+            ad = row.ad_group_ad.ad
+            print(f'Responsive search ad with resource name '
+                  f'"{ad.resource_name}", '
+                  f'status {aga_status_enum.Name(row.ad_group_ad.status)} '
+                  f'was found.')
+            if ad.responsive_search_ad:
```
devchas: If the search query includes `WHERE ad_group_ad.ad.type = RESPONSIVE_SEARCH_AD` why do we need this check here? If this check _is_ needed here, shouldn't the one_found flip to True be located under this check?

devchas: Just realized the answer to my own question - the type could be RSA but the RSAInfo may not be set.

PR #230, opened 2020-02-24
add_app_campaign: fixed multiple campaign criteria

wihl:
also split ad group and app ad creation and minor formatting.

Based on review comments in https://github.com/googleads/google-ads-php/pull/275

### DIFF
--- a/examples/advanced_operations/add_app_campaign.py
+++ b/examples/advanced_operations/add_app_campaign.py
@@ -34,19 +34,23 @@ from google.ads.google_ads.errors import GoogleAdsException
 def main(client, customer_id):
     """Main function for running this example."""
     try:
-        # Create the budget for the campaign.
+        # Creates the budget for the campaign.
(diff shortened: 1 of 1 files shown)


examples/advanced_operations/add_app_campaign.py
```diff
@@ -252,9 +258,22 @@ def _create_app_ad(client, customer_id, campaign_resource_name):
...
+    """Creates an App ad for a given ad group.
+
+    Args:
+        client: an initialized GoogleAdsClient instance.
+        customer_id: a client customer ID str.
+        ad_group_resource_name: the ad group where the ad will be added.
 
-    # Create the ad group ad
+    Returns:
+        None.
```
BenRKarl: We probably don't need this `Returns` doc, or the return statement at the end of the function, since it'll return `None` anyway. I know it's not part of the changes but we might as well clean it up here.

wihl: fixed.

PR #236, opened 2020-03-05
updates to get_responsive_search_ads example

wihl:
1) Added useful comment re: optional parameter

2) Took out check for responsive_search_ad_info and cleaned up.

### DIFF
--- a/examples/basic_operations/get_responsive_search_ads.py
+++ b/examples/basic_operations/get_responsive_search_ads.py
@@ -39,6 +39,8 @@ def main(client, customer_id, page_size, ad_group_id=None):
         WHERE ad_group_ad.ad.type = RESPONSIVE_SEARCH_AD
         AND ad_group_ad.status != "REMOVED"'''
 
+    # Optional: Specify an ad group ID to restrict search to only a given
+    # ad group.
     if ad_group_id:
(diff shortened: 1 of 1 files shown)


examples/basic_operations/get_responsive_search_ads.py
```diff
@@ -56,16 +58,11 @@ def main(client, customer_id, page_size, ad_group_id=None):
...
-                responsive_search_ad_info = ad.responsive_search_ad
-                print('Headlines:\n{}\n'
-                      'Descriptions:\n{}\n'.format(
-                          '\n'.join(_ad_text_assets_to_strs(
-                              client, responsive_search_ad_info.headlines)),
-                          '\n'.join(_ad_text_assets_to_strs(
-                              client, responsive_search_ad_info.descriptions))))
-            else:
-                print('\tResponsive search ad info was not found.')
+            print('Headlines:\n{}\nDescriptions:\n{}\n'.format(
```
BenRKarl: Any reason you use `format` here instead of an f-string?

wihl: Yes, with the `.join`, I think it would have been messier with an F-string

BenRKarl: Ok sounds good

PR #242, opened 2020-03-12
Defer loading of error_pb2 module until an exception is found in a response

BenRKarl:
Loading `_pb2` modules is a known bottleneck and this change defers the cost until the module is needed for exception handling.

### DIFF
--- a/google/ads/google_ads/interceptors/interceptor.py
+++ b/google/ads/google_ads/interceptors/interceptor.py
@@ -152,11 +152,11 @@ class Interceptor:
         return _ClientCallDetails(method, timeout, metadata, credentials)
 
     def __init__(self, api_version):
-        self._error_protos = import_module(
-            f'google.ads.google_ads.{api_version}.proto.errors.errors_pb2')
+        self._error_protos = None
         self._failure_key = (
             f'google.ads.googleads.{api_version}.errors.googleadsfailure-bin')
         self._exception = None
+        self._api_version = api_version
 
     def _get_error_from_response(self, response):
         """Attempts to wrap failed responses as GoogleAdsException instances.
@@ -231,6 +231,10 @@ class Interceptor:
             for kv in trailing_metadata:
                 if kv[0] == self._failure_key:
                     try:
+                        if not self._error_protos:
+                          self._error_protos = import_module(
+                            f'google.ads.google_ads.{self._api_version}.proto.'
+                            'errors.errors_pb2')
                         ga_failure = self._error_protos.GoogleAdsFailure()
                         ga_failure.ParseFromString(kv[1])
                         return ga_failure
--- a/tests/interceptors/interceptor_test.py
+++ b/tests/interceptors/interceptor_test.py
@@ -15,6 +15,7 @@
 
 
 from importlib import import_module
+import mock
 from unittest import TestCase
 
 from google.ads.google_ads.client import _DEFAULT_VERSION as default_version
@@ -88,3 +89,19 @@ class InterceptorTest(TestCase):
         interceptor = Interceptor(default_version)
         result = interceptor._get_google_ads_failure(None)
         self.assertEqual(result, None)
+
+    def test_init_no_error_proto_load(self):
+        """Ensures that error proto modules are not loaded on init."""
+        interceptor = Interceptor(default_version)
+        self.assertEqual(interceptor._error_protos, None)
+
+    def test_deferred_error_proto_module_load(self):
+        """Tests that import_module is called when an API error is received."""
+        with mock.patch('google.ads.google_ads.interceptors.'
+                        'interceptor.import_module') as import_mock:
+            interceptor = Interceptor(default_version)
+            mock_metadata = ((interceptor._failure_key, _MOCK_FAILURE_VALUE),)
+            interceptor._get_google_ads_failure(mock_metadata)
+            import_mock.assert_called_once_with('google.ads.google_ads.'
+                                                f'{default_version}.proto.'
+                                                'errors.errors_pb2')


google/ads/google_ads/interceptors/interceptor.py
```diff
@@ -231,6 +231,10 @@ def _get_google_ads_failure(self, trailing_metadata):
             for kv in trailing_metadata:
                 if kv[0] == self._failure_key:
                     try:
+                        if not self._error_protos:
+                          self._error_protos = import_module(
```
bobhancock: Do we want to wrap this in a try/except block? If the import fails we should fail elegantly with information on why we failed. Otherwise, I believe the next line will cause a stacktrace.

BenRKarl: Yes possibly - how do you think we should handle the failure case? Should this function return a different type of error? I also don't hate the idea of letting this raise a stack trace since a failure to load this module is not an expected error state.

wihl:
Similarly if put in an invalid refresh_token, I get:

```
Traceback (most recent call last):
  File "get_expanded_text_ads.py", line 68, in <module>
    google_ads_client = (google.ads.google_ads.client.GoogleAdsClient
  File "<REDACTED>/Projects/ads/v3/google-ads-python/google/ads/google_ads/client.py", line 159, in load_from_storage
    kwargs = cls._get_client_kwargs(config_data)
  File "<REDACTED>/Projects/ads/v3/google-ads-python/google/ads/google_ads/client.py", line 64, in _get_client_kwargs 
    return {'credentials': oauth2.get_credentials(config_data),
  File "<REDACTED>/Projects/ads/v3/google-ads-python/google/ads/google_ads/oauth2.py", line 96, in get_credentials
    config_data.get('refresh_token'))
  File "<REDACTED>/Projects/ads/v3/google-ads-python/google/ads/google_ads/oauth2.py", line 38, in initialize_credentials_wrapper
    credentials.refresh(Request())
  File "<REDACTED>/Projects/ads/v3/v3env/lib/python3.7/site-packages/google/oauth2/credentials.py", line 182, in refresh
    self._scopes,
  File "<REDACTED>/Projects/ads/v3/v3env/lib/python3.7/site-packages/google/oauth2/_client.py", line 248, in refresh_grant 
    response_data = _token_endpoint_request(request, token_uri, body)
  File "<REDACTED>/Projects/ads/v3/v3env/lib/python3.7/site-packages/google/oauth2/_client.py", line 124, in _token_endpoint_request
    _handle_error_response(response_body)
  File "<REDACTED>/Projects/ads/v3/v3env/lib/python3.7/site-packages/google/oauth2/_client.py", line 60, in _handle_error_response
    raise exceptions.RefreshError(error_details, response_body)
google.auth.exceptions.RefreshError: ('invalid_grant: Bad Request', '{\n  "error": "invalid_grant",\n  "error_description": "Bad Request"\n}')
```

wihl:
If I kill the TCP connect during a long stream, I get another bad error
(use 
`ss -K dst [2607:f8b0:4006:818::200a] dport = https` 
  (replace `2607:f8b0:4006:818::200a` by the current `nslookup` of `googleads.googleapis.com`))

```
Request made: ClientCustomerId: <REDACTED>, Host: googleads.googleapis.com:443, Method: /google.ads.googleads.v3.services.GoogleAdsService/SearchStream, RequestId: None, IsFault: True, FaultMessage: Software caused connection abort
Traceback (most recent call last):
  File "<REDACTED>/Projects/ads/v3/v3env/lib/python3.7/site-packages/google/api_core/grpc_helpers.py", line 79, in next
    return six.next(self._wrapped)
  File "<REDACTED>/Projects/ads/v3/google-ads-python/google/ads/google_ads/interceptors/exception_interceptor.py", line 100, in __next__ 
    raise e
  File "<REDACTED>/Projects/ads/v3/google-ads-python/google/ads/google_ads/interceptors/exception_interceptor.py", line 97, in __next__
    self._failure_handler(self._underlay_call)
  File "<REDACTED>/Projects/ads/v3/google-ads-python/google/ads/google_ads/interceptors/exception_interceptor.py", line 140, in _handle_grpc_failure
    raise self._get_error_from_response(response)
  File "<REDACTED>/Projects/ads/v3/google-ads-python/google/ads/google_ads/interceptors/exception_interceptor.py", line 92, in __next__
    return next(self._underlay_call)
  File "<REDACTED>/Projects/ads/v3/v3env/lib/python3.7/site-packages/grpc/_channel.py", line 416, in __next__ 
    return self._next()
  File "<REDACTED>/Projects/ads/v3/v3env/lib/python3.7/site-packages/grpc/_channel.py", line 689, in _next
    raise self
grpc._channel._MultiThreadedRendezvous: <_MultiThreadedRendezvous of RPC that terminated with:
        status = StatusCode.UNAVAILABLE
        details = "Software caused connection abort"
        debug_error_string = "{"created":"@1584050655.243204086","description":"Error received from peer ipv6:[2607:f8b0:4006:818::200a]:443","file":"src/core/lib/surface/call.cc","file_line":1056,"grpc_message":"Software caused connection abort","grpc_status":14}"
>

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "get_campaigns.py", line 65, in <module>
    main(google_ads_client, args.customer_id)
  File "get_campaigns.py", line 38, in main
    for batch in response:
  File "<REDACTED>/Projects/ads/v3/v3env/lib/python3.7/site-packages/google/api_core/grpc_helpers.py", line 81, in next
    six.raise_from(exceptions.from_grpc_error(exc), exc)
  File "<string>", line 3, in raise_from
google.api_core.exceptions.ServiceUnavailable: 503 Software caused connection abort
```

google/ads/google_ads/interceptors/interceptor.py
jradcliff: Does `import_module` raise an exception on failure? I'd certainly hope so.

BenRKarl: I'm leaning towards keeping this as-is. `import_module` does raise on failure, and we wouldn't really be able to provide a much better custom message than "could not load this error module."

AnashOommen: Rethrowing is the right thing to do, but I'm not sure how to surface the "could not load this error module". Can an error logging suffice? I mean, if we rethrow, the users will get an RpcException whereas they anticipate a GoogleAdsException, right? It will lead to failure of another form immediately after?

wihl: They will get a [ModuleNotFoundError](https://docs.python.org/3/library/exceptions.html#ModuleNotFoundError), which is more correct than rethrowing GoogleAdsException

wihl: I think we should **not** wrap and return a friendly message. Otherwise the customer will have trouble diagnosing the root cause. It also sets a bad precedent for catching any number of system errors, which is a bad practice.

Issue #245, opened 2020-03-16
ChangeStatus get/search failure using resource_name


j-walker23:
Hello, ran into a random issue, so I figured i'd pass it along.

Neither ChangeStatusService nor GoogleAdsService can get a resource using a resource_name from previous search results.

```
  service = client.get_service('ChangeStatusService')
  ga = client.get_service('GoogleAdsService')

  q = f"""
  select change_status.resource_name
  from change_status
  order by change_status.last_change_date_time desc
  limit 1
  """

  resource_name = None
  for row in ga.search(self.cid, query=q):
    resource_name = row.change_status.resource_name

  assert resource_name

  q = f"""
  select
    change_status.resource_status
  , change_status.resource_type
  , change_status.last_change_date_time
  from change_status
  where change_status.resource_name = '{resource_name}'
  """

  try:
    # Try GoogleAds
    list(ga.search(self.cid, query=q))
  except:
    # request_error: BAD_RESOURCE_ID
    pass

  try:
    # Try debugging only service get method
    list(service.get_change_status(resource_name))
  except:
    # request_error: BAD_RESOURCE_ID
    pass
```

![Error info here](https://user-images.githubusercontent.com/2018885/76735551-bcad9000-672a-11ea-9ee1-867c78dc95f5.png)

ps. That image shows a customer number, I can't think of a way that's unsafe, since auth would be required and i didn't show any tokens. But, please let me know if i'm wrong about that.

wihl:
The error is actually coming from the `list(service.get_change_status(resource_name))`. ChangeStatus returns a [ChangeStatus](https://developers.google.com/google-ads/api/reference/rpc/google.ads.googleads.v3.resources#google.ads.googleads.v3.resources.ChangeStatus) resource which is not iterable. Since you are coercing a non-iterable into a list, you are generating a `TypeError: 'ChangeStatus' object is not iterable` exception.

j-walker23:
Sorry, I shouldn’t have put pseudo code.
My code is heavily abstracted, and I forgot how to use it plainly for the
examples.

I included an image so you can see the error.

wihl:
Ok, I've been able to reproduce your issue with your specific customer ID and Ad Group ID. I'll file a bug.

j-walker23:
Perfect, thanks so much for your help as always!

BenRKarl:
@j-walker23 it seems that a fix for this issue may have been released, can you verify whether you're still having this problem?

[...]

PR #248, opened 2020-03-17
Fix Response Error Caching in Streaming Interceptors

BenRKarl:
When implementing streaming interceptors for v3 we moved the logic to parse and instantiate `GoogleAdsExceptions` into the `Interceptor` parent class because the logging interceptor for unary-streaming requests needs access to this functionality. This is for two reasons:

1. Unary-streaming interceptors are not executed in a guaranteed order.
2. There's no mechanism to pass exceptions between interceptors, like there is in unary-unary requests.

While implementing the shared logic we cached the exception on the parent class in an attempt to reduce the number of times it needed to be created. However that caused exceptions to be cached across requests, as a single Interceptor class is instantiated for each _service_.

This PR removes that caching mechanism and adds a test ensuring sequential calls to `Interceptor._get_error_from_response` do not return the same object.

### DIFF
--- a/google/ads/google_ads/interceptors/interceptor.py
+++ b/google/ads/google_ads/interceptors/interceptor.py
@@ -155,7 +155,6 @@ class Interceptor:
         self._error_protos = None
         self._failure_key = (
             f'google.ads.googleads.{api_version}.errors.googleadsfailure-bin')
-        self._exception = None
         self._api_version = api_version
 
     def _get_error_from_response(self, response):
@@ -182,9 +181,6 @@ class Interceptor:
             Exception: If not a GoogleAdsException or RpcException the error
                 will be raised as-is.
         """
-        if self._exception:
-            return self._exception
-
         status_code = response.code()
         response_exception = response.exception()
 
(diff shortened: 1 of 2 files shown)


wihl:
Actually this doesn't fix #245 which is specific to ChangeStatus. (Please update the description otherwise that issue may get inadvertently closed). This behavior is independent of ChangeStatus but still needs to be fixed.

google/ads/google_ads/interceptors/interceptor.py
```diff
@@ -200,20 +196,19 @@ def _get_error_from_response(self, response):
                 # library-specific Error type for easy handling. These errors
                 # originate from the Google Ads API and are often caused by
                 # invalid requests.
-                self._exception = GoogleAdsException(
+                return GoogleAdsException(
                     response_exception, response, google_ads_failure,
                     request_id)
             else:
                 # Raise the original exception if not a GoogleAdsFailure. This
```
bobhancock: nit: we are returning the exception, not raising/throwing it.

google/ads/google_ads/interceptors/interceptor.py
```diff
@@ -200,20 +196,19 @@ def _get_error_from_response(self, response):
...
                 # type of error is generally caused by problems at the request
                 # level, such as when an invalid endpoint is given.
-                self._exception = response_exception
+                return response_exception
         else:
             # Raise the original exception if error has status code
             # INTERNAL or RESOURCE_EXHAUSTED, meaning that
```
bobhancock: Are we missing text in the doc string? "meaning that" seems to be missing a clause.

BenRKarl:
I'm going to benchmark this change before merging it.

google/ads/google_ads/interceptors/interceptor.py
BenRKarl: Fixed.

google/ads/google_ads/interceptors/interceptor.py
BenRKarl: Fixed.

PR #252, opened 2020-03-24
add get_all_videos_and_images.py code example

xflesym:
add get_all_videos_and_images.py code example


examples/misc/get_all_videos_and_images.py
```diff
@@ -0,0 +1,68 @@
...
+    # Creates a query that will retrieve all video and image files.
+    query = ('SELECT media_file.id, media_file.name, media_file.type '
+             'FROM media_file ORDER BY media_file.id')
+
+    # Issues a search request by specifying page size.
+    results = ga_service.search(customer_id, query=query, page_size=page_size)
+
+    media_type_enum = client.get_type('MediaTypeEnum', version='v3').MediaType
+
+    try:
```
fiboknacky: As you also include many comments here, could you include one more here too?
"Iterates over all rows and prints the information about each media file."

xflesym: Yup sure. Added comment here.

PR #253, opened 2020-03-24
converted two old examples to use search_stream

wihl:
As part of doc update to streaming, moving over two old examples to streaming

### DIFF
--- a/examples/basic_operations/get_expanded_text_ads.py
+++ b/examples/basic_operations/get_expanded_text_ads.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright 2018 Google LLC
+# Copyright 2020 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -18,14 +18,14 @@
 import argparse
 import sys
 
-import google.ads.google_ads.client
+from google.ads.google_ads.client import GoogleAdsClient
+from google.ads.google_ads.errors import GoogleAdsException
 
 
-_DEFAULT_PAGE_SIZE = 1000
-
-
-def main(client, customer_id, page_size, ad_group_id=None):
+def main(client, customer_id, ad_group_id=None):
     ga_service = client.get_service('GoogleAdsService', version='v3')
+    ad_group_ad_status_enum = client.get_type(
+        'AdGroupAdStatusEnum', version='v3').AdGroupAdStatus
 
     query = ('SELECT ad_group.id, ad_group_ad.ad.id, '
              'ad_group_ad.ad.expanded_text_ad.headline_part1, '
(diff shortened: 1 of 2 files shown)


examples/billing/get_billing_setup.py
```diff
@@ -37,47 +35,41 @@ def main(client, customer_id, page_size):
...
+                billing_setup = row.billing_setup
+                payments_account_info = billing_setup.payments_account_info
+                print(f'Billing setup with ID {billing_setup.id.value}, '
+                      f'status "{billing_setup_status_enum.Name(billing_setup.status)}", '
+                      f'payments_account "{billing_setup.payments_account.value}" '
+                      f'payments_account_id "{payments_account_info.payments_account_id.value}", '
+                      f'payments_account_name "{payments_account_info.payments_account_name.value}", '
+                      f'payments_profile_id {payments_account_info.payments_profile_id.value}, '
+                      f'payments_profile_name "{payments_account_info.payments_profile_name.value}", '
+                      f'secondary_payments_profile_id "{payments_account_info.secondary_payments_profile_id.value}".')
```
BenRKarl: Nit: this is a _little_ tough to read, would it be worth setting some of these values to variables above the print statement so that the printed string is a easier to parse visually?

wihl: Used a shorter prefix variable name. Looking forward to getting rid of `.value`

examples/basic_operations/get_expanded_text_ads.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright 2018 Google LLC
+# Copyright 2020 Google LLC
```
fiboknacky: No, we don't change copy right for the existing examples.

examples/billing/get_billing_setup.py
```diff
@@ -37,47 +35,41 @@ def main(client, customer_id, page_size):
...
-        print('Request with ID "%s" failed with status "%s" and includes the '
-              'following errors:' % (ex.request_id, ex.error.code().name))
+        for batch in response:
+            for row in batch.results:
+                billing_setup = row.billing_setup
+                pai = billing_setup.payments_account_info
+                print(f'Billing setup with ID {billing_setup.id.value}, '
+                      f'status "{billing_setup_status_enum.Name(billing_setup.status)}", '
+                      f'payments_account "{billing_setup.payments_account.value}" '
+                      f'payments_account_id "{pai.payments_account_id.value}", '
```
fiboknacky: ID doesn't need quotes. Same for secondary_payments_profile_id below.

wihl: Note that if `secondary_payments_profile_id` is empty (as it is in my case), the result is just a blank space.

fiboknacky: That's why we have [this logic](https://github.com/googleads/google-ads-php/blob/master/examples/Billing/GetBillingSetup.php#L141) in PHP's. :)

examples/billing/get_billing_setup.py
```diff
@@ -37,47 +35,45 @@ def main(client, customer_id, page_size):
...
-        print('Request with ID "%s" failed with status "%s" and includes the '
-              'following errors:' % (ex.request_id, ex.error.code().name))
+        for batch in response:
+            for row in batch.results:
+                billing_setup = row.billing_setup
+                pai = billing_setup.payments_account_info
+                if pai.secondary_payments_profile_id.value:
+                    secondary_payments_profile_id = pai.secondary_payments_profile_id.value
+                else:
+                    secondary_payments_profile_id = "None"
```
fiboknacky: Can't this be just:

```python
secondary_payments_profile_id = (pai.secondary_payments_profile_id.value 
                                 if pai.secondary_payments_profile_id.value
                                 else "None")
```

?

wihl: It can be, but I'm not a big fan of that syntax as it hides intentionality somewhat. Since the rest of this example does not use it, I didn't want to introduce it.

PR #261, opened 2020-04-12
correct print format of generate_keyword_ideas.py

HassyHeisaku:
Print format correction to show "average monthly searches".


HassyHeisaku:
@googlebot I signed it!

HassyHeisaku:
@googlebot I fixed it.

HassyHeisaku:
Hi,
I changed to use "f-string".
And fixed cla/google problem.

HassyHeisaku:
@googlebot I fixed it.

HassyHeisaku:
@googlebot I fixed it.

BenRKarl:
@HassyHeisaku thanks for the changes! One more piece of feedback - we enforce an 80 character width limit, so lines `89` and `90` will need to be put on separate lines, I suggest something like this:

```python
            print(f'Keyword idea text "{idea.text.value}" has '                 
                     f'"{idea.keyword_idea_metrics.avg_monthly_searches.value}" '  
                     f'average monthly searches and "{competition_value}" '        
                     'competition.\n')
```

Also the print statement on line 94 should be separate like this:

```python
        print(f'Request with ID "{ex.request_id}" failed with status '
                  f'"{ex.error.code().name}" and includes the following errors:')
```

HassyHeisaku:
updated.

BenRKarl:
@HassyHeisaku sorry I think this branch got out of date, would you mind rebasing it against the current master branch? I'll try to get it merged afterwards as quickly as possible.

HassyHeisaku:
rebased and updated ;-)

PR #262, opened 2020-04-13
Update retrieval of customer ID

BenRKarl:
The `GetCustomerRequest` object sends a resource name instead of a customer ID so this updates the logging interceptor slightly to parse the customer ID from it. 

Mirroring change made in Ruby here: https://github.com/googleads/google-ads-ruby/pull/29/files

### DIFF
--- a/google/ads/google_ads/interceptors/logging_interceptor.py
+++ b/google/ads/google_ads/interceptors/logging_interceptor.py
@@ -122,8 +122,6 @@ class LoggingInterceptor(Interceptor, UnaryUnaryClientInterceptor,
         elif hasattr(request, 'resource_name'):
             return getattr(request, 'resource_name').split('/')[-1]
 
-        return 'N/A'
-
     def _parse_exception_to_str(self, exception):
         """Parses response exception object to str for logging.
 
--- a/google/ads/google_ads/interceptors/logging_interceptor.py
+++ b/google/ads/google_ads/interceptors/logging_interceptor.py
@@ -117,7 +117,12 @@ class LoggingInterceptor(Interceptor, UnaryUnaryClientInterceptor,
         Args:
             request: An instance of a request proto message.
         """
-        return getattr(request, 'customer_id', None)
+        if hasattr(request, 'customer_id'):
+            return getattr(request, 'customer_id')
+        elif hasattr(request, 'resource_name'):
+            return getattr(request, 'resource_name').split('/')[-1]
+
+        return 'N/A'
 
     def _parse_exception_to_str(self, exception):
         """Parses response exception object to str for logging.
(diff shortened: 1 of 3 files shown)


google/ads/google_ads/interceptors/logging_interceptor.py
```diff
@@ -117,7 +117,10 @@ def _get_customer_id(self, request):
         Args:
             request: An instance of a request proto message.
         """
-        return getattr(request, 'customer_id', None)
+        if hasattr(request, 'customer_id'):
+            return getattr(request, 'customer_id')
```
wihl, line 121: The Ruby version returns "N/A" (which is probably wrong since '/' is the delimiter for resource name parts). In Python, perhaps you should a default value if none is found, ie. "NONE" or "UNKNOWN"

BenRKarl: I think we thought about how to do this a long time ago when logging was first implemented and went with `None` since it will serialize into the string `'None'` when printed. In any case I don't think it was arbitrary to return `None` here initially and I'd prefer to stick with that.

wihl: SGTM

google/ads/google_ads/interceptors/logging_interceptor.py
```diff
@@ -117,7 +117,10 @@ def _get_customer_id(self, request):
...
+        elif hasattr(request, 'resource_name'):
+            return getattr(request, 'resource_name').split('/')[-1]
```
AnashOommen, line 123: This needs an additional check to see if the resource_name indeed starts with customer_id.

This logic will fail for languageConstants/{criterionId}, for instance,
Perhaps the Ruby code also needs some fix.

AnashOommen: You can refer to https://github.com/googleads/google-ads-dotnet/blob/master/src/V3/ResourceNames.cs, which has a manually curated list of all the known resource names.

BenRKarl: @AnashOommen the issue isn't really about the structure of the resource name, right? It's that the customer ID happens to be in a resource name on a ` GetCustomerRequest` instance? Maybe it's better to check the type of the request object instead of parsing the resource name string? Do we know if there are any other request objects that don't have a `customer_id` field?

BenRKarl: Talked with @AnashOommen offline, the idea here is just to do some light validation of the string to make sure it starts with `'customers'` so we can be sure it's a resource_name, otherwise return `None`. No need for a complete list, also type checking doesn't quite make sense because an arbitrary string can be set by the user on `GetCustomerRequest`. I'll push up a change.

google/ads/google_ads/interceptors/logging_interceptor.py
```diff
@@ -120,7 +120,10 @@ def _get_customer_id(self, request):
         if hasattr(request, 'customer_id'):
             return getattr(request, 'customer_id')
         elif hasattr(request, 'resource_name'):
-            return getattr(request, 'resource_name').split('/')[-1]
+            resource_name = getattr(request, 'resource_name')
+            segments = resource_name.split('/')
+            if segments[0] == 'customers':
+                return segments[1]
```
wihl, line 126: Might be good to have an explicit
```
else
    return None
```

Issue #260, opened 2020-04-10
Star imports from `v2/__init__.py` and `v3/__init__.py` fail


henribru:
`from google.ads.google_ads.v2 import *` and `from google.ads.google_ads.v3 import *` fail with `AttributeError: unknown sub-module 'Enums'.` This happens because `__all__` is defined to contain uppercase versions of all the keys in `_lazy_name_to_package_map`. This works for the services since there is a corresponding uppercase type for each snake case module, but the dict also contains `types` and `enums`, for which this is not the case.

BenRKarl:
@henribru we don't intend for the modules under each version to be imported directly, those objects should be retrieved using the `get_type` and `get_service` client methods. You can see usages throughout our [examples](https://github.com/googleads/google-ads-python/tree/master/examples). All the code under the version module is generated, and it's possible that the structure could change, which is why we have the getter methods to abstract that structure away.

Is there a specific use-case you have that necessitates `import *`? We could consider improvements to the client to allow retrieval of multiple classes at once.

henribru:
@BenRKarl I don't really need it, no, this is just something I noticed while working on type stubs for the library. But it's clearly a bug to put something in `__all__` that's not actually importable (it defeats the purpose of `__all__`), so I'm not sure how you can consider it working as intended.

BenRKarl:
@henribru Good point. I looked more closely at these files and the `_lazy_name_to_package_map` maps classnames to filenames and module paths so you can retrieve a class like:

```
from google.ads.google_ads.v3.services.google_ads_service import GoogleAdsService
```

However this pattern doesn't apply to `enums` and `types` as in there is no class at this path:

```
from google.ads.google_ads.v3.services.enums import Enums
```

So I changed this to a bug and will be posting a PR shortly, thanks for pointing it out.

BenRKarl:
This is fixed in `5.0.3` of the library.

PR #263, opened 2020-04-13
Update versioned __init__ file to isolate enums and types

BenRKarl:
`enums` and `types` don't fit the same pattern as the other classnames/files in the map, where a capitalized classname has a corresponding snakecase file name, i.e. `GoogleAdsService` and `google_ads_service` so this removes them from the map and exposes them via separate functionality. 

Resolves: https://github.com/googleads/google-ads-python/issues/260

### DIFF
--- a/google/ads/google_ads/v2/__init__.py
+++ b/google/ads/google_ads/v2/__init__.py
@@ -25,7 +25,6 @@ if sys.version_info < (3, 7):
 
 
 _lazy_name_to_package_map = dict(
-    types='google.ads.google_ads.v2',
     account_budget_proposal_service_client='google.ads.google_ads.v2.services',
     account_budget_service_client='google.ads.google_ads.v2.services',
     ad_group_ad_asset_view_service_client='google.ads.google_ads.v2.services',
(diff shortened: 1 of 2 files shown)


google/ads/google_ads/v2/__init__.py
```diff
@@ -265,6 +263,14 @@ def __getattr__(name):  # Requires Python >= 3.7
                      {'__doc__': sub_mod_class.__doc__, 'enums': enums})
         globals()[name] = klass
         return klass
+    elif name == 'enums':
+        module = importlib.import_module('google.ads.google_ads.v3.services.enums')
```
wihl, line 267: This is the v2 init so why are you importing the v3 enums?

google/ads/google_ads/v2/__init__.py
```diff
@@ -265,6 +263,14 @@ def __getattr__(name):  # Requires Python >= 3.7
...
+        globals()[name] = module
+        return module
+    elif name == 'types':
+        module = importlib.import_module('google.ads.google_ads.v3.types')
```
wihl, line 271: ditto previous comment

google/ads/google_ads/v2/__init__.py
BenRKarl: Ah just a typo, fixed

google/ads/google_ads/v2/__init__.py
BenRKarl: Fixed

Issue #267, opened 2020-04-20
Problem with `TargetRestrictionOperation` and `FeedAttributeOperation` enums


henribru:
There seems to still be a few problems with the `enums.py` files after the recent fix. Specifically, dynamically creating the enums for `TargetRestrictionOperation` and `FeedAttributeOperation` in the v3 version fail because the name of the inner class in the string is different from the class attribute they're stored in (triggering an assertion error in `_CreateEnumTypeUponFirstAccess.__getattribute__`), while in the v2 version `FeedAttributeOperation` fails because of incorrect indentation inside the string.

BenRKarl:
@henribru thanks for pointing this out - how are you triggering these errors? By importing the enum directly or by using `get_type`?

henribru:
I'm importing them directly. As far as I can tell you can't actually use `get_type` to import the enums defined in `enums.py`, as `get_type` imports from `types.py` and the `getattr` there imports enums from the protobuf files, not `enums.py`.

```python
>>> from google.ads.google_ads.v3.services.enums import TargetRestrictionOperation, FeedAttributeOperation
>>> TargetRestrictionOperation.TargetRestrictionOperation
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/henbruas/.cache/pypoetry/virtualenvs/google-ads-stubs-ACofHr-C-py3.7/lib/python3.7/site-packages/google/ads/google_ads/v3/services/enums.py", line 81, in __getattribute__
    assert class_def_src.startswith('class %s(enum.' % enum_type_name), 'Expected ' + enum_type_name
AssertionError: Expected TargetRestrictionOperation
>>> FeedAttributeOperation.FeedAttributeOperation
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/henbruas/.cache/pypoetry/virtualenvs/google-ads-stubs-ACofHr-C-py3.7/lib/python3.7/site-packages/google/ads/google_ads/v3/services/enums.py", line 81, in __getattribute__
    assert class_def_src.startswith('class %s(enum.' % enum_type_name), 'Expected ' + enum_type_name
AssertionError: Expected FeedAttributeOperation
>>> from google.ads.google_ads.v2.services.enums import FeedAttributeOperation
>>> FeedAttributeOperation.Operator
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #268, opened 2020-04-23
Fix issues with `services/enums.py`

BenRKarl:
Add test enforcing `service_client.enums` interface.
Remove non-v2 Enums from `v2/services/enums.py`
Update validation in lazy-loaded enums to check for inner Enum: `Operator`.

Fixes: https://github.com/googleads/google-ads-python/issues/267

### DIFF
--- a/google/ads/google_ads/v2/services/enums.py
+++ b/google/ads/google_ads/v2/services/enums.py
@@ -79,15 +79,15 @@ class Thing(enum.IntEnum):
                 # another must have already executed it and created our type.
                 return class_def_src
 
-            # Here `class_def_src` will be the stringified definition of the
+            # Here 'class_def_src' will be the stringified definition of the
             # inner enum. This assertion checks that the string starts with
-            # `'class AccessRole(enum.'` if, for example, the outer class name
+            # 'class AccessRole(enum.' if, for example, the outer class name
             # is AccessRoleEnum. This follows the pattern for the majority of
             # our API enums, where the outer message is FooBarEnum and the inner
             # enum is named FooBar. In a few cases enums are defined on
-            # Operations, for example `FeedAttributeOperation`, and the inner
-            # enum is always called `Operator` so this assertion also allows the
-            # pattern `'class Operator(enum.'`.
+            # Operations, for example 'FeedAttributeOperation', and the inner
+            # enum is always called 'Operator' so this assertion also allows the
+            # pattern 'class Operator(enum.'.
             assertion = class_def_src.startswith(
                 'class %s(enum.' % enum_type_name) or class_def_src.startswith(
                     'class Operator(enum.')
--- a/google/ads/google_ads/v2/services/enums.py
+++ b/google/ads/google_ads/v2/services/enums.py
@@ -78,6 +78,16 @@ class Thing(enum.IntEnum):
                 # If the class attribute has been replaced with something else,
                 # another must have already executed it and created our type.
                 return class_def_src
+
+            # Here `class_def_src` will be the stringified definition of the
+            # inner enum. This assertion checks that the string starts with
+            # `'class AccessRole(enum.'` if, for example, the outer class name
+            # is AccessRoleEnum. This follows the pattern for the majority of
+            # our API enums, where the outer message is FooBarEnum and the inner
+            # enum is named FooBar. In a few cases enums are defined on
+            # Operations, for example `FeedAttributeOperation`, and the inner
+            # enum is always called `Operator` so this assertion also allows the
+            # pattern `'class Operator(enum.'`.
             assertion = class_def_src.startswith(
                 'class %s(enum.' % enum_type_name) or class_def_src.startswith(
                     'class Operator(enum.')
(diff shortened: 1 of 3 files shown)


google/ads/google_ads/v2/services/enums.py
```diff
@@ -78,7 +78,10 @@ def __getattribute__(self, name):
                 # If the class attribute has been replaced with something else,
                 # another must have already executed it and created our type.
                 return class_def_src
-            assert class_def_src.startswith('class %s(enum.' % enum_type_name), 'Expected ' + enum_type_name
+            assertion = class_def_src.startswith(
+                'class %s(enum.' % enum_type_name) or class_def_src.startswith(
+                    'class Operator(enum.')
+            assert assertion, 'Expected ' + enum_type_name
```
BenRKarl, line 84: `class_def_src` will be the stringified definition of the enum, for example [here](https://github.com/googleads/google-ads-python/blob/2e11de5f0edbb561b6c186aeef95a5605336723a/google/ads/google_ads/v3/services/enums.py#L163). This assertion checks that the string either starts with `'class AccessRole(enum.'` if, for example, the outer class name is `AccessRoleEnum`. This follows the pattern for the majority of our API enums, (see [here](https://github.com/googleapis/googleapis/blob/master/google/ads/googleads/v3/enums/access_role.proto#L31)), where the outer message is `FooBarEnum` and the inner enum is named `FooBar`. In a few cases enums are defined on Operations, for example [`FeedAttributeOperation`](https://github.com/googleapis/googleapis/blob/bf839ae632e0f263a729569e44be4b38b1c85f9c/google/ads/googleads/v1/resources/feed.proto#L175), and the inner enum is called `Operator` so this assertion also allows the patter `'class Operator(enum.'`

tests/client_test.py
```diff
@@ -472,3 +474,60 @@ def test_init_with_logging_config(self):
             Client.GoogleAdsClient(mock_credentials_instance,
                                    self.developer_token, logging_config=config)
             mock_dictConfig.assert_called_once_with(config)
+
+    def test_service_client_dot_enums_property(self):
```
BenRKarl, line 478: This test is a little hard to follow, but basically it tests the `service.enum` property for a service in every version of the API, which implicitly tests the `services/enums.py` file, which exposes the enums used by the property. It loops over each member of `service.enum` and compares it to the raw enum pb2 file found by calling `client.get_type` with the enum's name.

google/ads/google_ads/v2/services/enums.py
wihl: This is valuable information so therefore should be in the comments rather than left in this PR.

google/ads/google_ads/v2/services/enums.py
```diff
@@ -5444,19 +5373,19 @@ class ExternalConversionSource(enum.IntEnum):
...
+          UNSPECIFIED (int): Unspecified.
+          UNKNOWN (int): Used for return value only. Represents value unknown in this version.
+          ADD (int): Add the attribute to the existing attributes.
+        """
+        UNSPECIFIED = 0
+        UNKNOWN = 1
+        ADD = 2
 '''
-FeedAttributeOperation = FeedAttributeOperation()
+FeedAttributeOperation = FeedAttributeOperation() # For __getattribute__
```
wihl, line 5388: could you either expand or remove this comment?

google/ads/google_ads/v3/services/enums.py
```diff
@@ -78,7 +78,10 @@ def __getattribute__(self, name):
                 # If the class attribute has been replaced with something else,
                 # another must have already executed it and created our type.
                 return class_def_src
-            assert class_def_src.startswith('class %s(enum.' % enum_type_name), 'Expected ' + enum_type_name
+            assertion = class_def_src.startswith(
+                'class %s(enum.' % enum_type_name) or class_def_src.startswith(
+                    'class Operator(enum.')
+            assert assertion, 'Expected ' + enum_type_name
```
wihl, line 84: Perhaps expand assertion to include both enum and operator cases?

tests/client_test.py
wihl: Again, this comment belongs in the code rather than in the PR.

google/ads/google_ads/v2/services/enums.py
BenRKarl: Sounds good I'll add.

tests/client_test.py
BenRKarl: Ok will add.

google/ads/google_ads/v2/services/enums.py
BenRKarl: This is just for consistency, since it's next to all of the other corresponding lines in this file.

google/ads/google_ads/v3/services/enums.py
BenRKarl: Do you mean separate it into two assertions and assert that either one is true?

wihl: No, I mean to expand the assertion message to cover the `operator` case as well. The current message could be ambiguous if I understand correctly.

BenRKarl: Oh I see - I'm not sure that would add much, this is just asserting that the name of the first attribute on the class is of a certain pattern, either `FooBar` for the class `FooBarEnum` or simply `Operator`. In either case it's akin to an `AttributeError`.

BenRKarl:
@wihl updated the backticks to apostrophes

PR #270, opened 2020-05-06
Update README for usability. 

BenRKarl:
All documentation will be consolidated on devsite at: https://developers.google.com/google-ads/api/docs/client-libs/python


README.rst
```diff
@@ -13,213 +13,10 @@ Features
 * Easy management of credentials.
 * Easy creation of Google Ads API service clients.
 
-Requirements
-------------
-* Python 3.7+
```
mcloonan, line 18: You can keep this section. Quick reference info that is useful at a glance can be useful in the GitHub repo.

BenRKarl: Sounds good, in that case I think I'll add the one-line installation instructions too because that's very useful to have here.

wihl:
Approval will be on hold pending v3_1 release.

BenRKarl:
Once this is merged let's remove the wiki, since it also has a bunch of redundant information. It's removed via Settings.

README.rst
```diff
@@ -7,219 +7,26 @@ Build Status
...
 Installation
-############
-
-This library is distributed via PyPI. If you have not already done so, install
-`pip`_, the following command to install this client library:
-
+------------
 .. code-block::
 
   pip install google-ads
```
wihl, line 18: do we even need this line when there is more detailed install info coming to devsite?

BenRKarl: We decided to leave quick reference info like this here for people who just want to know the installation command. I had initially taken it out but put it back in after talking with @mcloonan

PR #302, opened 2020-09-09
Added UploadCallConversion example

AndrewMBurke:


examples/remarketing/upload_call_conversion.py
```diff
@@ -0,0 +1,174 @@
...
+        type=str,
+        required=True,
+        help="The the date and time of the conversion (should be after the "
+        "click time). The format is 'yyyy-mm-dd hh:mm:ss+|-hh:mm', e.g. "
+        "'2019-01-01 12:32:45-08:00'.",
+    )
+    parser.add_argument(
+        "-v",
+        "--conversion_value",
+        type=str,
```
wihl, line 160: not sure if a string is most appropriate here. The value should be a currency value (https://support.google.com/google-ads/answer/3419241?hl=en) so let the parser filter out bad data like characters.

examples/remarketing/upload_call_conversion.py
```diff
@@ -0,0 +1,174 @@
...
+        caller_id: The caller ID from which this call was placed. Caller ID is
+            expected to be in E.164 format with preceding '+' sign,
+            e.g. '+16502531234'.
+        call_start_date_time: The date and time at which the call occurred. The
+            format is 'yyyy-mm-dd hh:mm:ss+|-hh:mm',
+            e.g. '2019-01-01 12:32:45-08:00'.
+        conversion_date_time: The the date and time of the conversion (should be
+            after the click time). The format is 'yyyy-mm-dd hh:mm:ss+|-hh:mm',
+            e.g. '2019-01-01 12:32:45-08:00'.
+        conversion_value: The conversion value.
```
wihl, line 52: Should be a currency value (see comment below)

examples/remarketing/upload_call_conversion.py
```diff
@@ -0,0 +1,174 @@
...
+
+    # Create a call conversion in USD currency.
+    call_conversion = client.get_type("CallConversion", version="v5")
+    call_conversion.conversion_action = client.get_service(
+        "ConversionActionService", version="v5"
+    ).conversion_action_path(customer_id, conversion_action_id)
+    call_conversion.caller_id = caller_id
+    call_conversion.call_start_date_time = call_start_date_time
+    call_conversion.conversion_date_time = conversion_date_time
+    call_conversion.conversion_value = float(conversion_value)
```
wihl, line 67: if you filter at the argparse level, you should remove the cast

examples/remarketing/upload_call_conversion.py
AndrewMBurke: Fixed.

examples/remarketing/upload_call_conversion.py
AndrewMBurke: Done.

wihl (approved):

BenRKarl (approved):

PR #307, opened 2020-09-11
Add `add_customer_negative_crierion` example.

BenRKarl:

PR #308, opened 2020-09-11
Add `add_geo_target` example.

BenRKarl:


examples/extensions/add_geo_target.py
```diff
@@ -0,0 +1,115 @@
...
+        "ExtensionFeedItemService", version="v5"
+    )
+
+    extension_feed_item_op = client.get_type(
+        "ExtensionFeedItemOperation", version="v5"
+    )
+    extension_feed_item = extension_feed_item_op.update
+    # Creates an extension feed item using the specified feed item ID and
+    # geo target constant ID for targeting.
+    extension_feed_item.resource_name = extension_feed_item_service.extension_feed_item_path(
```
BenRKarl, line 46: The auto-formatter forces this onto a single line, even if I manually newline it with a leading `(`. I could set the resource_name to a separate variable if it's worth staying under the char limit.

AndrewMBurke: I'm OK with an exception to this line's length, especially given that it's only the result of two long, descriptive names. I feel like I've seen a couple of other examples experience a similar issue, though can't recall at the moment.

BenRKarl: Yeah some of the names in the API are just very long, and using something like `ext_feed_it_serv` would be horribly difficult to read.

examples/extensions/add_geo_target.py
```diff
@@ -0,0 +1,115 @@
...
+        customer_id: a client customer ID.
+        feed_item_id: the ID of an extension feed item.
+        geo_target_constant_id: the geo target constant ID to add to the
+            extension feed item.
+    """
+    extension_feed_item_service = client.get_service(
+        "ExtensionFeedItemService", version="v5"
+    )
+
+    extension_feed_item_op = client.get_type(
```
AndrewMBurke, line 40: Nit: almost all the examples use `operation` instead of `op`.

examples/extensions/add_geo_target.py
```diff
@@ -0,0 +1,115 @@
...
+# limitations under the License.
+"""Adds a geo target to an extension feed item for targeting."""
+
+
+import argparse
+import sys
+
+from google.ads.google_ads.client import GoogleAdsClient
+from google.ads.google_ads.errors import GoogleAdsException
+from google.api_core import protobuf_helpers
```
AndrewMBurke, line 23: Import `protobuf_helpers` before Google Ads items.

AndrewMBurke (approved):
LGTM with a couple of very minor comments.

examples/extensions/add_geo_target.py
wihl: Like in PR #307, you could break at the `.`

wihl (approved):
LGTM with noted comments

examples/extensions/add_geo_target.py
BenRKarl: The auto-formatter will set it back to how it is now. I'd prefer not to fight it that hard. The alternative seems to rename the variables but I'm not sure there's a way to do that without sacrificing readability.

examples/extensions/add_geo_target.py
BenRKarl: SG updated.

examples/extensions/add_geo_target.py
BenRKarl: Oh my 😅  Thanks! fixed.

PR #311, opened 2020-09-12
Add `remove_feed_item` example

BenRKarl:


examples/feeds/remove_feed_items.py
```diff
@@ -0,0 +1,107 @@
...
+        )
+        operations.append(feed_item_operation)
+
+    try:
+        response = feed_item_service.mutate_feed_items(customer_id, operations)
+
+        for feed_item in response.results:
+            print(
+                f"Removed feed item with resource name: "
+                "'{feed_item.resource_name}'"
```
AndrewMBurke, line 56: F-string out of place, should be:
```
"Removed feed item with resource name: "
f"'{feed_item.resource_name}'"
```

AndrewMBurke (approved):
LGTM with one minor fix.

examples/feeds/remove_feed_items.py
```diff
@@ -0,0 +1,107 @@
...
+    feed_item_service = client.get_service("FeedItemService", version="v5")
+
+    operations = []
+    for feed_item_id in feed_item_ids:
+        # Constructs an operation that will remove the feed item based on the
+        # resource name.
+        feed_item_operation = client.get_type("FeedItemOperation", version="v5")
+        feed_item_operation.remove = feed_item_service.feed_item_path(
+            customer_id,
+            # constructs a resource name for a feed_item, which is in the
```
wihl, line 44: nit: s/constructs/Constructs

examples/feeds/remove_feed_items.py
```diff
@@ -0,0 +1,107 @@
...
+        "-c",
+        "--customer_id",
+        type=str,
+        required=True,
+        help="The Google Ads customer ID.",
+    )
+    parser.add_argument(
+        "-f",
+        "--feed_id",
+        type=str,
```
wihl, line 90: I wonder if we should make the feed_id and the feed_items_ids as integers insteads of strs so as to enforce better typing at the arg level

examples/feeds/remove_feed_items.py
wihl, line 44: This comment needs to be moved up two lines as `.feed_item_path` creates the full resource name whereas `.format_composite` only creates `{feed_id}~{feed_item_id}`

wihl (approved):
LGTM with minor comments and suggestions

examples/feeds/remove_feed_items.py
BenRKarl: Done

examples/feeds/remove_feed_items.py
BenRKarl: Let's discuss separately in the context of the other examples and other libraries.

examples/feeds/remove_feed_items.py
BenRKarl: Done

PR #340, opened 2020-11-23
Fix syntax error in add_sitelink.py example

BenRKarl:
This line wasn't tested previously because it only executes if the current date is after Nov. 20.


AndrewMBurke (approved):

wihl (approved):
LGTM

PR #344, opened 2020-12-01
Fix bug in add_lead_form_extension example.

BenRKarl:
Adding a duplicate `PHONE_NUMBER` LeadFormField caused an error in our function test framework.


wihl (approved):

laurachevalier4 (approved):
lgtm

PR #345, opened 2020-12-01
Fix bug in add_listing_scope example.

BenRKarl:


wihl (approved):
Unfortunate that there are a few of things lingering around.

PR #351, opened 2020-12-07
create_feed_item_set example

laurachevalier4:
Main thing I'm looking for feedback on is the location extension / affiliate extension code. Without a GMB account, I wasn't able to test this (but I did verify that the protos are properly set)

### DIFF
--- a/examples/feeds/create_feed_item_set.py
+++ b/examples/feeds/create_feed_item_set.py
@@ -64,8 +64,8 @@ def main(client, customer_id, feed_id):
         "FeedItemSetStringFilterTypeEnum", version="v6"
     ).EXACT
     """
-    """
     # 2) Affiliate Extension
+    """
     dynamic_affiliate_location_set_filter = feed_item_set.dynamic_affiliate_location_set_filter
     dynamic_affiliate_location_set_filter.chain_ids.extend([INSERT_CHAIN_IDS_HERE])
     """
--- a/examples/feeds/create_feed_item_set.py
+++ b/examples/feeds/create_feed_item_set.py
@@ -49,26 +49,23 @@ def main(client, customer_id, feed_id):
     feed_item_set.feed = feed_resource_name
     feed_item_set.display_name = "Feed Item Set #%s" % uuid.uuid4()
 
-    """A feed item set can be created as a dynamic set by setting an optional
-    filter field below. If your feed is a location extension, uncomment the code
-    that sets dynamic_location_set_filter. If your feed is an affiliate
-    extension, set dynamic_affiliate_location_set_filter instead.
-    """
+    # A feed item set can be created as a dynamic set by setting an optional
+    # filter field below. If your feed is a location extension, uncomment the code
+    # that sets dynamic_location_set_filter. If your feed is an affiliate
+    # extension, set dynamic_affiliate_location_set_filter instead.
+
     # 1) Location Extension
     # NOTE: Does not support Curated location extensions.
-    """
-    dynamic_location_set_filter = feed_item_set.dynamic_location_set_filter
-    business_name_filter = dynamic_location_set_filter.business_name_filter
-    business_name_filter.business_name = 'INSERT_YOUR_BUSINESS_NAME_HERE'
-    business_name_filter.filter_type = client.get_type(
-        "FeedItemSetStringFilterTypeEnum", version="v6"
-    ).EXACT
-    """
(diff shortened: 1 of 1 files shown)


examples/feeds/create_feed_item_set.py
```diff
@@ -0,0 +1,117 @@
...
+    feed_item_set.feed = feed_resource_name
+    feed_item_set.display_name = "Feed Item Set #%s" % uuid.uuid4()
+
+    """A feed item set can be created as a dynamic set by setting an optional
+    filter field below. If your feed is a location extension, uncomment the code
+    that sets dynamic_location_set_filter. If your feed is an affiliate
+    extension, set dynamic_affiliate_location_set_filter instead.
+    """
+    # 1) Location Extension
+    # NOTE: Does not support Curated location extensions.
```
laurachevalier4, line 58: I tried to test this by creating a location extension using Curated locations, since I do not have a Google My Business account, and received error: "The feed type of the parent Feed is not compatible with the type of dynamic filter being set." According to the [docs](https://developers.google.com/google-ads/api/reference/rpc/v6/FeedItemSetErrorEnum.FeedItemSetError), this happens when the feed item set type isn't location-based, e.g. "you can only set dynamic_location_set_filter for LOCATION feed item sets." I would think this would work for location feeds from Google My Business but haven't found a way to verify (that doesn't involve creating a GMB account and plenty of mock data).

examples/feeds/create_feed_item_set.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+
+    feed_item_set = feed_item_set_operation.create
+    feed_resource_name = client.get_service(
+        "FeedService", version="v6"
+    ).feed_path(customer_id, feed_id)
+    feed_item_set.feed = feed_resource_name
+    feed_item_set.display_name = "Feed Item Set #%s" % uuid.uuid4()
+
+    """A feed item set can be created as a dynamic set by setting an optional
```
wihl, line 52: It is confusing mixing multi-line comments followed by # comments followed by multiline commented out blocks of code. I suggest you use only # comments in this code block per https://google.github.io/styleguide/pyguide.html#385-block-and-inline-comments

examples/feeds/create_feed_item_set.py
wihl: Tested

examples/feeds/create_feed_item_set.py
```diff
@@ -0,0 +1,117 @@
...
+    # The following argument(s) should be provided to run the example.
+    parser.add_argument(
+        "-c",
+        "--customer_id",
+        type=str,
+        required=True,
+        help="The Google Ads customer ID.",
+    )
+    parser.add_argument(
+        "-f", "--feed_id", type=str, required=True, help="The feed ID.",
```
wihl, line 111: use consistent formatting for argument arguments

examples/feeds/create_feed_item_set.py
BenRKarl: Agree, and same applies for the lines below as well.

examples/feeds/create_feed_item_set.py
```diff
@@ -0,0 +1,117 @@
...
+    feed_item_set_operation = client.get_type(
+        "FeedItemSetOperation", version="v6"
+    )
+
+    feed_item_set = feed_item_set_operation.create
+    feed_resource_name = client.get_service(
+        "FeedService", version="v6"
+    ).feed_path(customer_id, feed_id)
+    feed_item_set.feed = feed_resource_name
+    feed_item_set.display_name = "Feed Item Set #%s" % uuid.uuid4()
```
BenRKarl, line 50: Use an f-string here.

BenRKarl (changes requested):

examples/feeds/create_feed_item_set.py
laurachevalier4: Ack. I was following the pattern set in the Golden but it makes sense that commenting practices wouldn't be consistent across our libraries. I'll make sure to consult the styleguide next time :)

examples/feeds/create_feed_item_set.py
laurachevalier4: done

examples/feeds/create_feed_item_set.py
laurachevalier4: done

BenRKarl (approved):

wihl (approved):

PR #354, opened 2020-12-10
remove_entire_sitelink_campaign_extension_setting example

laurachevalier4:


examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
```diff
@@ -0,0 +1,230 @@
...
+    # Iterate through each row and append the extension feed item resource
+    # names to the return array.
+    for batch in response:
+        for row in batch.results:
+            extension_feed_item_resource_names += (
+                row.campaign_extension_setting.extension_feed_items
+            )
+
+    if len(extension_feed_item_resource_names) == 0:
+        raise Exception(
```
wihl, line 171: should specify the type of exception more precisely

examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
```diff
@@ -0,0 +1,230 @@
...
+
+    Returns:
+        An array of MutateOperations for the extension feed items.
+    """
+    mutate_operations = []
+    # Create a MutateOperation for each extension feed item to remove.
+    for resource_name in extension_feed_item_resource_names:
+        mutate_operation = client.get_type("MutateOperation", version="v6")
+        mutate_operation.extension_feed_item_operation.remove = resource_name
+        mutate_operations.append(mutate_operation)
```
wihl, line 198: why are you mixing `<list>.append()` and `<list> += <value>` in the same file?

laurachevalier4: I used += to concat a list, and append for a single item

examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
```diff
@@ -0,0 +1,230 @@
...
+        client: an initialized GoogleAdsClient instance.
+        customer_id: a client customer ID.
+        campaign_id: the campaign ID.
+    """
+    # Initialize an array of MutateOperations
+    mutate_operations = []
+    sitelink_campaign_extension_setting_mutate_operation = _create_sitelink_campaign_extension_setting_mutate_operation(
+        client, customer_id, campaign_id
+    )
+    mutate_operations += [sitelink_campaign_extension_setting_mutate_operation]
```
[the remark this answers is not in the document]
laurachevalier4: I thought it would be more clear if I initialized it as empty first. I can do this instead though

examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
laurachevalier4: Do you think `ValueError` makes sense here?

examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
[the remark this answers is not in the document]
wihl: your call, not a biggie either way

examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
wihl: yes

examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
BenRKarl: We would normally use `extend` in the other examples, maybe go with that for consistency?

examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
```diff
@@ -0,0 +1,230 @@
...
+
+    ga_service = client.get_service("GoogleAdsService", version="v6")
+    extension_feed_item_resource_names = _get_all_sitelink_extension_feed_items(
+        client, ga_service, customer_id, campaign_id
+    )
+    extension_feed_item_mutate_operations = _create_extension_feed_item_mutate_operations(
+        client, extension_feed_item_resource_names
+    )
+    mutate_operations += extension_feed_item_mutate_operations
```
BenRKarl, line 59: Maybe use `extend` here for consistency.

examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
wihl: personally if I was using `.append()`, I'd use `.extend()` but also not a big deal

examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
```diff
@@ -0,0 +1,230 @@
...
+        print(
+            "Removed a campaign extension setting with resource name "
+            f'"{mutate_operation_responses[0].campaign_extension_setting_result.resource_name}".'
+        )
+        # Iterate through the remainder of the mutate operation responses,
+        # which are the extension feed item removal responses
+        for mutate_operation_response in mutate_operation_responses[1:]:
+            print(
+                "Removed an extension feed item with resource name "
+                f'"{mutate_operation_response.extension_feed_item_result}"'
```
BenRKarl, line 75: This should access the `extenstion_feed_item_result.resource_name` field.

examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
```diff
@@ -0,0 +1,230 @@
...
+
+    Args:
+        client: an initialized GoogleAdsClient instance.
+        ga_service: the Google Ads API service client.
+        customer_id: the client customer ID.
+        campaign_id: the ID of the campaign to get the sitelink extension feed
+            items from.
+
+    Returns:
+        An array of str resource names of extension feed items.
```
wihl, line 138: should mention Raises in the PyDoc per https://google.github.io/styleguide/pyguide.html#244-decision

examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
```diff
@@ -0,0 +1,230 @@
...
+    extension_type_name = extension_type_enum.ExtensionType.Name(
+        extension_type_enum.ExtensionType.SITELINK
+    )
+
+    # Construct the query.
+    query = f"""
+        SELECT
+            campaign_extension_setting.campaign,
+            campaign_extension_setting.extension_type,
+            campaign_extension_setting.extension_feed_items
```
BenRKarl, line 153: These should only be indented 2 spaces not 4, since we use SQL style for these strings.

examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
BenRKarl: Does the golden raise an exception? I'm kind of curious why if so, because nothing will catch it. Why not just print a message about why the example can't complete and exit with 0?

BenRKarl (changes requested):

examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
laurachevalier4: the golden raises an InvalidArgumentException. The whole runExample method in the golden is within a try [block](https://github.com/googleads/google-ads-php/blob/8a8bd8efbae248c3f80cdadacc7ca10a38b6cd8e/examples/Extensions/RemoveEntireSitelinkCampaignExtensionSetting.php#L73)

laurachevalier4: I think @BenRKarl's suggestion makes sense. So how about in the `if` block:
```
print("The specified campaign does not contain a sitelink campaign extension setting.")
sys.exit(0)
```

examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
laurachevalier4: I will do this if I don't end up going with printing and exiting

examples/extensions/remove_entire_sitelink_campaign_extension_setting.py
```diff
@@ -0,0 +1,231 @@
...
+        client: an initialized GoogleAdsClient instance.
+        customer_id: a client customer ID.
+        campaign_id: the campaign ID.
+    """
+    # Initialize an array of MutateOperations
+    mutate_operations = []
+    sitelink_campaign_extension_setting_mutate_operation = _create_sitelink_campaign_extension_setting_mutate_operation(
+        client, customer_id, campaign_id
+    )
+    mutate_operations.extend([sitelink_campaign_extension_setting_mutate_operation])
```
wihl, line 50: why coerce `sitelink_campaign_extension_setting_mutate_operation` into a list when you can just use `.append` in this case?

laurachevalier4: No good reason at all. Fixed!

wihl (approved):

BenRKarl (approved):
Just change the error code from 0 to 1 and this is all set.

PR #355, opened 2020-12-10
Added the get_feed_items_of_feed_item_set example.

fiboknacky:


examples/feeds/get_feed_items_of_feed_item_set.py
```diff
@@ -0,0 +1,96 @@
...
+import argparse
+import sys
+
+from google.ads.google_ads.client import GoogleAdsClient
+from google.ads.google_ads.errors import GoogleAdsException
+from google.ads.google_ads.util import ResourceName
+
+
+def main(client, customer_id, feed_id, feed_item_set_id):
+    ga_service = client.get_service("GoogleAdsService", version="v6")
```
wihl, line 31: Need PyDoc comment string

examples/feeds/get_feed_items_of_feed_item_set.py
```diff
@@ -0,0 +1,96 @@
...
+    feed_item_set_service = client.get_service("FeedItemSetService",
+                                               version="v6")
+
+    query = ("SELECT feed_item_set_link.feed_item FROM feed_item_set_link "
```
wihl, line 35: current convention is to use multiline f-string with SQLstyle formatting

examples/feeds/get_feed_items_of_feed_item_set.py
```diff
@@ -0,0 +1,96 @@
...
+
+    # Issues a search request using streaming.
+    response = ga_service.search_stream(customer_id, query=query)
+
+    print("The feed items with the following resource names are linked with "
+          f"the feed item set with ID {feed_item_set_id}:")
+    try:
+        for batch in response:
+            for row in batch.results:
+              print(f"'{row.feed_item_set_link.feed_item}'")
```
wihl, line 49: optional: there is no delimiter between each value. Did you want them each on a separate line? If so, maybe indent with a `\t`?

examples/feeds/get_feed_items_of_feed_item_set.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def main(client, customer_id, feed_id, feed_item_set_id):
+    ga_service = client.get_service("GoogleAdsService", version="v6")
+    feed_item_set_service = client.get_service("FeedItemSetService",
+                                               version="v6")
+
+    query = ("SELECT feed_item_set_link.feed_item FROM feed_item_set_link "
+             "WHERE feed_item_set_link.feed_item_set = '%s'"
+             % feed_item_set_service.feed_item_set_path(
```
BenRKarl, line 37: Use f-strings instead of the `%` operator for interpolation. And it might be easier to set this to its own variable than to make the call to format_composite inside the f-string brackets.

examples/feeds/get_feed_items_of_feed_item_set.py
```diff
@@ -0,0 +1,96 @@
...
+                 customer_id, ResourceName.format_composite(
+                     feed_id, feed_item_set_id)))
+
+    # Issues a search request using streaming.
+    response = ga_service.search_stream(customer_id, query=query)
```
BenRKarl, line 42: Does the golden specify a page_size here? Usually we're explicit about these.

BenRKarl (changes requested):

examples/feeds/get_feed_items_of_feed_item_set.py
fiboknacky: Alright. Note that get_campaigns doesn't have this PyDoc.

wihl: Yes, we haven't updated all examples to use the new conventions yet. (We did update the internal style documentation).

examples/feeds/get_feed_items_of_feed_item_set.py
```diff
@@ -0,0 +1,96 @@
...
+import sys
+
+from google.ads.google_ads.client import GoogleAdsClient
+from google.ads.google_ads.errors import GoogleAdsException
+from google.ads.google_ads.util import ResourceName
+
+
+def main(client, customer_id, feed_id, feed_item_set_id):
+    ga_service = client.get_service("GoogleAdsService", version="v6")
+    feed_item_set_service = client.get_service("FeedItemSetService",
```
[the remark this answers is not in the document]
fiboknacky: Done.

examples/feeds/get_feed_items_of_feed_item_set.py
fiboknacky: Yes, I avoided using f-strings at first because of the function calling.
I'll create a new variable instead.

examples/feeds/get_feed_items_of_feed_item_set.py
```diff
@@ -0,0 +1,96 @@
...
+                                               version="v6")
+
+    query = ("SELECT feed_item_set_link.feed_item FROM feed_item_set_link "
+             "WHERE feed_item_set_link.feed_item_set = '%s'"
```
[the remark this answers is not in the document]
fiboknacky: Done.

examples/feeds/get_feed_items_of_feed_item_set.py
fiboknacky: It's a search stream, so no page_size specified in the golden CL.

examples/feeds/get_feed_items_of_feed_item_set.py
fiboknacky: It's already aligned with the golden CL.

examples/feeds/get_feed_items_of_feed_item_set.py
```diff
@@ -0,0 +1,108 @@
...
+    # The following argument(s) should be provided to run the example.
+    parser.add_argument(
+        "-c",
+        "--customer_id",
+        type=str,
+        required=True,
+        help="The Google Ads customer ID.",
+    )
+    parser.add_argument(
+        "-i", "--feed_id", type=str, required=True, help="The Feed ID.",
```
wihl, line 95: should format like the other arguments. Black is not perfect.

wihl (approved):
LGTM with minor formatting comment (no need for another review)

BenRKarl (approved):

examples/feeds/get_feed_items_of_feed_item_set.py
BenRKarl: Even if this is formatted manually `black` will set it back to one line, so I think we should leave this as-is.

PR #359, opened 2020-12-15
Fix param name in add_hotel_ads example

BenRKarl:


wihl (approved):
LGTM. As long as you're in the file, maybe fix the exception handler to be the new style?

PR #368, opened 2021-01-15
Release version 8.2.0

BenRKarl:


wihl (approved):
Presumably the new environment variables will be documented at https://developers.google.com/google-ads/api/docs/client-libs/python/configuration in the near future

PR #373, opened 2021-01-19
Fix issues in add_hotel_listing_group_tree example

BenRKarl:
This example wasn't correctly creating entities to cover the "others" case, as is done [here](https://github.com/googleads/google-ads-php/blob/master/examples/HotelAds/AddHotelListingGroupTree.php#L260) in the PHP example.

### DIFF
--- a/examples/hotel_ads/add_hotel_listing_group_tree.py
+++ b/examples/hotel_ads/add_hotel_listing_group_tree.py
@@ -245,6 +245,12 @@ def _add_level1_nodes(
     other_hotels_listing_dimension_info = client.get_type(
         "ListingDimensionInfo", version="v6"
     )
+    # This is necessary to set "hotel_class" as the oneof field on the
+    # ListingDimentionInfo object without specifying the optional
+    # hotel_class.value field.
+    other_hotels_listing_dimension_info.hotel_class.CopyFrom(
+        client.get_type("HotelClassInfo", version="v6")
+    )
 
     # Create listing group info for other hotel classes as a SUBDIVISION node,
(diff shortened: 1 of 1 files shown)


examples/hotel_ads/add_hotel_listing_group_tree.py
```diff
@@ -245,6 +245,12 @@ def _add_level1_nodes(
     other_hotels_listing_dimension_info = client.get_type(
         "ListingDimensionInfo", version="v6"
     )
+    # This is necessary to set "hotel_class" as the oneof field on the
```
wihl, line 248: I don't think that "This is necessary" is necessary :)

examples/hotel_ads/add_hotel_listing_group_tree.py
```diff
@@ -351,6 +357,12 @@ def _add_level2_nodes(
     other_hotel_regions_listing_dimension_info = client.get_type(
         "ListingDimensionInfo", version="v6"
     )
+    # This is necessary to set "hotel_country_region" as the oneof field on the
```
wihl, line 360: ditto prior comment

examples/hotel_ads/add_hotel_listing_group_tree.py
BenRKarl: Done.

examples/hotel_ads/add_hotel_listing_group_tree.py
BenRKarl: Done.

wihl (approved):

PR #379, opened 2021-01-27
Fixes #364: Treat customer_client.id as str

### LINKED ISSUE #364: trouble running get_account_hierarchy.py example?
Hi I'm trying to run the get_account_heirarchy.py example and I keep running into this error:
Traceback (most recent call last):
  File ".\get_account_hierarchy.py", line 202, in <module>
    main(google_ads_client, args.login_customer_id)
  File ".\get_account_hierarchy.py", line 101, in main
    customer_client.id
  File "C:\Python\Python38\lib\site-packages\google\protobuf\message.py", line 94, in __hash__
    raise TypeError('unhashable object')
TypeError: unhashable object

I'm running this command from the project directory:
python .\get_account_hierarchy.py -l 0123456789

Also, I've tried:
python .\get_account_hierarchy.py --manager_customer_id 0123456789

(the argument is actually our manager account customer id and I do have the yaml file with our creds in the directory)

Could you please tell me what I'm doing wrong?

arammaliachi:
As described in #364, 

Currently script runs into a `TypeError` for trying to find `customer_client.id` of type `<class 'google.protobuf.wrappers_pb2.Int64Value'>` in the keys of a dictionary.  Keys must be strings so the following if block runs into an error:

```
if (
    customer_client.id
    not in customer_ids_to_child_accounts
    and customer_client.level == 1
)
```

`customer_client.id` must be within a  **str**. Actually source code got it right at [line 173](https://github.com/googleads/google-ads-python/blob/d90d9e9aadfa3b81a95110f806547010c0e0824d/examples/account_management/get_account_hierarchy.py#L173)

### DIFF
--- a/examples/account_management/get_account_hierarchy.py
+++ b/examples/account_management/get_account_hierarchy.py
@@ -120,7 +120,7 @@ def main(client, login_customer_id=None):
                         # prevent visiting the same customer many times, we
                         # need to check if it's already in the Dictionary.
                         if (
-                            customer_client.id
+                            str(customer_client.id)
                             not in customer_ids_to_child_accounts
                             and customer_client.level == 1
                         ):


arammaliachi:
@googlebot I signed it!

BenRKarl:
@arammaliachi thanks for catching this - the `id` value, as of v5 I think, should be an integer and not an `Int64Value` since we dropped support for wrapper types. There's still an issue though, that the given `login_customer_id` is treated as a string, but other IDs retrieved from the API are ints, so there could be a mismatch when searching the dict. 

Since the IDs we get from the API are already ints, perhaps instead of converting the ints to strings, we should just convert the passed in login_customer_id to an int [here](https://github.com/googleads/google-ads-python/pull/379/files#diff-d96bac34a7188a1d5c80f35dfc4a5c9e03c5f2ec7d4ae7006c650fb01ad886deR203). 

What do you think?

arammaliachi:
I see what you're saying @BenRKarl with code as is the `customer_ids_to_child_accounts` dictionary ended up having both strings and ints as keys which was also causing the printing recursive function to only go down 1 level.

I submitted new PR #394 with changes that should fix both the dictionary and the recursive printing function. During a personal clean-up I deleted my original fork sorry for the overhead, we can close this PR if thats ok.

BenRKarl:
Closing this in favor of #394

PR #401, opened 2021-03-31
Fix inconsistent region tags

PierrickVoulet:


examples/hotel_ads/add_hotel_ad.py
```diff
@@ -171,7 +171,7 @@ def _add_hotel_campaign(
     # Set the campaign network options. Only Google Search is allowed for hotel
     # campaigns.
     campaign.network_settings.target_google_search = True
-    # [END add_hotel_ad]
+    # [END add_hotel_ad_1]
```
laurachevalier4, line 174: If these `add_hotel_ad_1 ` tags are wrapped by the `add_hotel_ad` tags, are they going to be visible to the user? Or are those removed nested tags also removed?

PierrickVoulet: No, all lines that define region tags are removed even is embedded in another one.

laurachevalier4 (approved):

BenRKarl (approved):
Thank you @PierrickVoulet

PR #413, opened 2021-04-22
Update README.rst

batardo:
Hey Guys, 

We upgraded to your new google ads. We formally used google Adwords api. We were using python Python 3.6.9 and got stuck while calling get_service method after initiating our client:

image

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -9,7 +9,7 @@ Build Status
 
 Requirements
 ------------
-* Python 3.6+
+* Python 3.7+
 
 Installation
 ------------


wihl:
Python v3.6 is still supported by [this line](https://github.com/googleads/google-ads-python/blob/a39748521847e85138fca593f3be2681352ad024/setup.py#L55) and we intend to continue to support it. If Python 3.6.x is not working, please [file an issue](https://github.com/googleads/google-ads-python/issues/new/choose) and we'll take a look.

PR #449, opened 2021-06-15
Changes for release v8_0.

laurachevalier4:


wihl (approved):

google/ads/googleads/v8/__init__.py
```diff
@@ -0,0 +1,1711 @@
...
+        return klass
+    else:
+        raise AttributeError(f"unknown type {name!r}.")
+
+
+def __dir__():
+    return globals().get("__all__") or __getattr__("__all__")
+
+
+if not sys.version_info >= (3, 7):
```
BenRKarl, line 1708: Can you change this to `(3, 6)`?

BenRKarl (changes requested):

google/ads/googleads/v8/__init__.py
BenRKarl: This is good as-is actually!

BenRKarl (approved):

PR #465, opened 2021-07-07
Release version 13.0.0

BenRKarl:


laurachevalier4 (approved):

ChangeLog
```diff
@@ -1,3 +1,10 @@
+* 13.0.0
+- Deprecate support for Python 3.6
```
AnashOommen, line 2: I think this should be something on the lines of "Bumped the minimum Python version requirement to Python 3.7+. As per https://www.python.org/dev/peps/pep-0494/, Python 3.6 will reach EOL in Dec 2021. Users are encouraged to upgrade to this version of the library. If you need more time to upgrade from Python 3.6, you can continue to use v12.0.0, which will be supported until the sunset of Google Ads API v8".

AnashOommen (changes requested):
I added a comment, but we need Thanet's guidance to be sure on the v8 sunset timeline.

ChangeLog
BenRKarl: Connected offline and added these details to the README instead.

AnashOommen (approved):

PR #473, opened 2021-07-28
Expand `google-api-core pin` to `>1, <3`

busunkim96:
To avoid diamond dependency conflicts with Cloud libraries, the Ads library should keep a `>1, <3` pin on google-api-core until https://github.com/googleapis/google-cloud-python/issues/10566 is closed.

See discussion in [doc](https://docs.google.com/document/d/1euAvUsia_4zf98lNvpwA3K0o2b4y5Xr9xAkyzCnIMzQ/edit#).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,10 @@ import io
 
 install_requires = [
     "google-auth-oauthlib >= 0.3.0, < 1.0.0",
-    "google-api-core >= 1.21.0, < 2.0.0",
+    # NOTE: Maintainers, please do not require google-api-core>=2.x.x
+    # Until this issue is closed
+    # https://github.com/googleapis/google-cloud-python/issues/10566
+    "google-api-core >= 1.21.0, < 3.0.0",
     "googleapis-common-protos >= 1.5.8, < 2.0.0",
     "grpcio >= 1.38.1, < 2.0.0",
     "proto-plus >= 1.18.0, < 2.0.0",


AnashOommen:
I see the doc, but it is unclear what the consequences of expanding this to
3.0.0 is.

busunkim96:
Ah sorry, here's some more context:

`google-api-core` is planning on officially dropping 2.7 in a new major release (2.0.0).

The particular situation this PR is looking to prevent is one where some customer who uses both`google-ads` and a Google Cloud library sees a diamond dependency conflict.  As an example, if some recent release of `google-ads` wants `google-api-core>1.x.x, <2.0.0dev` and some recent release of `google-cloud-bigquery` wants `google-api-core>2.x.x, <3.0.0dev` pip will not be able to find a version of `google-api-core` that satisfies both and raise an error. 

On the Cloud side, we're doing this range expansion before the `google-api-core` release to increase the amount of time / # of versions published with compatible pins. The hope is that this decreases the likelihood end users of these libraries see dependency conflicts.

If you all would be more comfortable waiting until `google-api-core` 2.0.0 is released to take this change that is also OK.

AnashOommen:
Hi Bu,

I read the document and understood the diamond dependency issue. That LGTM
in terms of the proposed approach.

We have a very large scale migration going on right now. Any bugs /
performance regressions introduced in google-api-core 2.0.0 will have a
much worse outcome than the diamond dependency issue we are talking about
here. Happy to share details offline.

Out of abundance of caution, our recommendation is to kindly pause your
release until we run this new library version internally against our
internal benchmarks and tests to ensure that nothing blows up.

Kindly start an internal email thread with me,  and Ben Karl, feel free to
cc Dan/Justin. We will loop in more folks as needed and will try to unblock
you ASAP.

busunkim96:
Hi Anash! 

One more important note: there are no major bug fixes/ performance changes planned for google-api-core 2.0.0, it is intended to just be a removal of python 2.7. You can review the currently open PR: https://github.com/googleapis/python-api-core/pull/212. 

I've also started an email thread to go into details.

BenRKarl:
Closing this because we're going to submit it via our internal code review process.

PR #487, opened 2021-08-31
Update sample google-ads.yaml file to illustrate the `use_proto_plus` flag

BenRKarl:


wihl (commented):
This file is fine, but please also update the links in README.RST which are broken, notably "Version 14.0.0_"

google-ads.yaml
```diff
@@ -7,6 +7,16 @@
 ###############################################################################
 developer_token: INSERT_DEVELOPER_TOKEN_HERE
 
+# Use proto plus
+###############################################################################
+# This parameter specifies whether the client library should return           #
+# proto-plus messages or protobuf messages. This value should be explicitly   #
+# set to either "True" or "False", For more information on the                #
+# differences between these two types, see our Protobuf Messages guide:       #
+# https://developers.google.com/google-ads/api/docs/client-libs/python/protobuf-messages #
```
laurachevalier4, line 16: Nit: expand the right edge of the table so it's as wide as the link (unless that's more anti-pep-8?)

wihl (approved):

google-ads.yaml
BenRKarl: Since this isn't a Python file I don't think it is subject to pep-8, so I updated it

BenRKarl:
@wihl I fixed the README link

laurachevalier4 (dismissed):

wihl (approved):

PR #491, opened 2021-09-03
Added the location criterion

fblascogarma:
I added the location criterion so the smart campaign that is being created has set the location target. Before, the location was only used to get budget recommendations. Now, the location will also be used as the location target for the campaign.


fblascogarma:
@googlebot I signed it!

fblascogarma:
@googlebot I signed it!

BenRKarl (approved):
This looks great, thanks for adding this Francisco. Once this gets merged I'll add an item to our next release so that it gets added to the other language examples.

PR #508, opened 2021-09-29
Release version 14.0.1

BenRKarl:
* Pin protobuf to < 3.18.0
* Add examples add_bidding_data_exclusion, add_bidding_seasonality_adjustment
* Style updates for examples that use search_stream
* Fix hardcoded URLs for media in various examples

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,16 +18,14 @@ import io
 
 install_requires = [
     "google-auth-oauthlib >= 0.3.0, < 1.0.0",
-    # NOTE: Require google-api-core > 1.31.2 until
-    # https://github.com/googleapis/google-cloud-python/issues/10566
-    # is resolved. Once resolved, require google-api-core >=2.x.x
-    "google-api-core >= 1.31.2, < 3.0.0",
+    "google-api-core >= 2.0.1, < 3.0.0",
     "googleapis-common-protos >= 1.5.8, < 2.0.0",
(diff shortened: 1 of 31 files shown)


setup.py
```diff
@@ -18,16 +18,14 @@
...
-    # is resolved. Once resolved, require google-api-core >=2.x.x
-    "google-api-core >= 1.31.2, < 3.0.0",
+    "google-api-core >= 2.0.1, < 3.0.0",
     "googleapis-common-protos >= 1.5.8, < 2.0.0",
     "grpcio >= 1.38.1, < 2.0.0",
     "proto-plus == 1.18.1",
     "PyYAML >= 5.1, < 6.0",
     "setuptools >= 40.3.0",
-    "nox >= 2020.12.31, < 2022.6",
+    "nox == 2020.12.31",
```
BenRKarl, line 27: @laurachevalier4 is this correct? I think I may want to revert this line, can you confirm?

laurachevalier4: I think so, yes. We didn't want to pin a particular version but there was no "upper bound" that we could predict so we set an arbitrary future date when we might expect breaking changes to be introduced.

wihl (approved):

setup.py
BenRKarl: I thought so, fixed this so it's the same as what's in the main branch now.

laurachevalier4 (approved):

Issue #441, opened 2021-05-22
Get Account Hierarchy Example - PERMISSION_DENIED Errors on Inaccessible Accounts/MCs With Depth > 1


davedavis:
I'm working on a project where I have to get the account list for a complex and deep hierarchy (using all 6 levels) and it's determined by the logged in user, so cannot be set manually (no manager ID can be specified). 

When running the the get_account_hierarchy example, it works as expected until it reaches a sub-MCC that contains accounts that are inaccessible by the user. 

In the screenshot below, you can see the example is initialized and the **top level accessible** customers are listed. 
The first accessible customer (91) is a directly linked account, not an MCC, so it prints correctly.

The second accessible customer (93) contains a whole bunch of directly managed accounts and sub MCCs. 
The IDs are all returned from the customer service get_customer() method. 

The problem occurs when one of those IDs returned is a sub manager account with accounts that are inaccessible to the user. In the below screenshot, it's querying the manager account ending (17) and getting a permission denied error.

![image](https://user-images.githubusercontent.com/5341256/119241797-155b9280-bb51-11eb-9a41-325b31c7bbc5.png)

You can see that this manager account, accessible to the user, contains two accounts that are inaccessible to the user

![image](https://user-images.githubusercontent.com/5341256/119242002-6324ca80-bb52-11eb-88aa-fe7e7c7bb1ad.png)

![image](https://user-images.githubusercontent.com/5341256/119242035-98311d00-bb52-11eb-878c-b88ca207f24f.png)

The user does not have access to any accounts in that Manager account. So, when it's queried, they cause the error (I think). 

So my question is, how can I bypass/skip these? I can't seem to figure out a way to differentiate between these as the customer_client object doesn't contain a field is exactly the same as all the rest (manager: True, Level: 1 etc.). Do I have to handle this situation by catching the exception or is there a more elegant way to do it? 

An example request ID if needed is: zApa0OQuxgx9gmrhr2kkxw

Many thanks in advance.

laurachevalier4:
Hey @davedavis! To help us understand what's happening here so that we can better advise you, can you mail your client-side logs to googleadsapi-support@google.com, include the login customer ID, and call out Anash (@AnashOommen) so that we can have a look? Thanks!

davedavis:
Sure @laurachevalier4. Appreciate the reply. 

I think this is specifically related to the example in this library though.

[...]

PR #522, opened 2021-10-23
fix in get_account_hierarchy

CefBoud:
I had the same error  mentionned in [#441 ](https://github.com/googleads/google-ads-python/issues/441) using code based on `examples/account_management/get_account_hierarchy.py`. 
 
> Error with message "User doesn't have permission to access customer. Note: If you're accessing a client customer, the manager's customer id must be set in the 'login-customer-id' header. See https://developers.google.com/google-ads/api/docs/concepts/call-structure#cid".

The error occurs in the `client.get_service("GoogleAdsService").search` method because the client doesn't have a `login_customer_id `.

Assigning the seed_customer_id (which is part of the accessible customers) to `GoogleAdsClient` that generates `GoogleAdsService` worked for me.
```python
for seed_customer_id in seed_customer_ids:
        client.login_customer_id = str(seed_customer_id)
        googleads_service = client.get_service("GoogleAdsService")

```


CefBoud:
@googlebot I signed it!

examples/account_management/get_account_hierarchy.py
```diff
@@ -88,6 +88,11 @@ def main(client, login_customer_id=None):
         unprocessed_customer_ids = [seed_customer_id]
         customer_ids_to_child_accounts = dict()
         root_customer_client = None
+        
+        # login_customer_id is required to query child accounts that are not
+        # directly accessible to the OAuth Credentials
+        client.login_customer_id = str(seed_customer_id)
+        googleads_service = client.get_service("GoogleAdsService")
```
BenRKarl, line 95: Seems like the first `googleads_service` initialization on line 43 can be removed, then line 79 can be updated to use customer service: 

```py

customer_id = customer_service.parse_customer_path(
    customer_resource_name
)["customer_id"]
```

BenRKarl (changes requested):

ThatsJustCheesy:
The changes here fixed the example for me. Previously, I was getting "User doesn't have permission to access customer."

The final patched code:

```py
for seed_customer_id in seed_customer_ids:
    client.login_customer_id = seed_customer_id
    googleads_service = client.get_service("GoogleAdsService")
```

using @BenRKarl's suggested changes

PR #554, opened 2021-12-21
Added the creation of listing group that is necessary for the PMax for retail campaign to run

fblascogarma:


AnashOommen (approved):

examples/shopping_ads/add_performance_max_retail_campaign.py
```diff
@@ -245,16 +245,20 @@ def _create_performance_max_campaign_operation(
     # Setting a portfolio bidding strategy by resource name is not supported.
     # Max Conversion and Max Conversion Value are the only strategies supported
     # for Performance Max campaigns.
+    campaign.bidding_strategy_type = (
+        client.enums.BiddingStrategyTypeEnum.MAXIMIZE_CONVERSION_VALUE
+    )
```
BenRKarl, line 250: I actually don't think this need to be set at all, we should delete these three lines.

examples/shopping_ads/add_performance_max_retail_campaign.py
```diff
@@ -263,11 +267,14 @@ def _create_performance_max_campaign_operation(
     # Set the Final URL expansion opt out. This flag is specific to
     # Performance Max campaigns. If opted out (True), only the final URLs in
     # the asset group or URLs specified in the advertiser's Google Merchant
-    # Center or business data feeds are targeted.
+    # Center or business data feeds are targeted. This is the recommended
+    # way for accounts that use their Merchant Center feed. As this is an
+    # example of PMax for Retail that uses a Merchant Center feed, we will
+    # set True the URL expansion opt out.
```
BenRKarl, line 273: Just so that this example is consistent with the .NET one, can you remove this additional text, and add the below right above the line where `url_expansion_opt_out` is set?

```
# For a Retail campaign, we want the final URL's to be limited to those explicitly surfaced via GMC.
```

examples/shopping_ads/add_performance_max_retail_campaign.py
```diff
@@ -439,6 +446,22 @@ def _create_asset_group_operation(
...
+    # https://developers.google.com/google-ads/api/reference/rpc/v9/AssetGroupListingGroupFilter
+    mutate_operation = client.get_type("MutateOperation")
+    asset_group_listing_group = mutate_operation.asset_group_listing_group_filter_operation.create
+    asset_group_listing_group.asset_group = asset_group_service.asset_group_path(
+            customer_id,
+            _ASSET_GROUP_TEMPORARY_ID,
+        )
+    asset_group_listing_group.type_ = client.enums.ListingGroupFilterTypeEnum.UNIT_INCLUDED
+    # Here are other options for ListingGroupFilterTypeEnum
+    # https://developers.google.com/google-ads/api/reference/rpc/v9/ListingGroupFilterTypeEnum.ListingGroupFilterType
```
BenRKarl, line 460: Can you replace the above comments with the below:

```
# Because this is a Performance Max campaign for retail, we need to specify that this is in the shopping vertical.
```

examples/shopping_ads/add_performance_max_retail_campaign.py
fblascogarma: that's correct, Ben. I tested this and the pmax campaign was created successfully with max conv value as the bidding strategy. I will delete these 3 lines, and add a new commented out line showing how to correctly set up the campaign if you want max conv instead of max conv value.

examples/shopping_ads/add_performance_max_retail_campaign.py
```diff
@@ -245,16 +245,20 @@ def _create_performance_max_campaign_operation(
...
     # A target_roas of 3.5 corresponds to a 350% return on ad spend.
-    campaign.bidding_strategy_type = (
-        client.enums.BiddingStrategyTypeEnum.MAXIMIZE_CONVERSION_VALUE
-    )
-    campaign.maximize_conversion_value.target_roas = 3.5
+    # campaign.maximize_conversion_value.target_roas = 3.5
+    # For first time users, it's recommended not to set a target ROAS.
+    # Although target ROAS is optional, you still need to define it
+    # even if you do not want to use it.
+    campaign.maximize_conversion_value.target_roas = None
```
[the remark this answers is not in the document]
fblascogarma: Google recommends for new accounts to not set a target ROAS at the beginning. Only when the advertiser knows how the campaign can perform. Therefore, the default None is more practical, and you need to define it because if not you will not be able to create the campaign as it is a required field.

examples/shopping_ads/add_performance_max_retail_campaign.py
fblascogarma: will do!

examples/shopping_ads/add_performance_max_retail_campaign.py
```diff
@@ -439,6 +446,22 @@ def _create_asset_group_operation(
     )
     operations.append(mutate_operation)
 
+    # Create a ListingGroupFilter and link it to the AssetGroup.
+    # This is necessary for the campaign to run.
+    # https://developers.google.com/google-ads/api/reference/rpc/v9/AssetGroupListingGroupFilter
```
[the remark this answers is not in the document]
fblascogarma: will do!

examples/shopping_ads/add_performance_max_retail_campaign.py
fblascogarma: will do!

examples/shopping_ads/add_performance_max_retail_campaign.py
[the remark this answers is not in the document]
BenRKarl: Ok this makes sense - I'll pass this along to the other lib owners so they can implement this.

BenRKarl (approved):

wihl (approved):
Thanks so much for doing this!

BenRKarl:
@fblascogarma this is ready to be merged, would you please rebase it so it's up-to-date with `main`?

PR #594, opened 2022-03-04
chg: adapt printout: total `clicks`/`impressions`

JakobSteixner:
The printed messages declare `impression`/`clicks` metrics to be *daily* impressions/clicks. This does not appear to be correct: When creating a keyword plan with the code example in `./add_keyword_plan.py` and generating metrics for that plan, the output values scale linearly with the timespan used in the `add_keyword_plan._create_keyword_plan` - i.e. roughly 4 times higher for `NEXT_MONTH` than for  `NEXT_WEEK`, roughly 3 times higher still for `NEXT_QUARTER`


BenRKarl (approved):

BenRKarl:
Verified also that the [PHP library](https://github.com/googleads/google-ads-php/blob/main/examples/Planning/GenerateForecastMetrics.php#L128) and [Java library](https://github.com/googleads/google-ads-java/blob/main/google-ads-examples/src/main/java/com/google/ads/googleads/examples/planning/GenerateForecastMetrics.java#L107) use `daily` in their log statement.

Issue #597, opened 2022-03-12
GoogleAdsService doesn't properly apply retries when Retry is passed in


Kache:
**Describe the bug:**

`GoogleAdsService` provides `search()`, which accepts a [`retry: google.api_core.retry.Retry`](https://github.com/googleapis/python-api-core/blob/c89f55d10df50c222d05ea046e07e05ef7d17725/google/api_core/retry.py#L223). The Retry decorator ends up only getting [applied on the first request](https://github.com/googleads/google-ads-python/blob/0b928474b588422f421aadf12948538378fc770f/google/ads/googleads/v10/services/services/google_ads_service/client.py#L3292-L3294), making it effectively useless for large results with many pages.

Subsequent requests are issued by `SearchPager`, [initialized a few lines down](https://github.com/googleads/google-ads-python/blob/0b928474b588422f421aadf12948538378fc770f/google/ads/googleads/v10/services/services/google_ads_service/client.py#L3298-L3300), where `retry` is not passed in, thus making it unavailable [at evocation](https://github.com/googleads/google-ads-python/blob/main/google/ads/googleads/v10/services/services/google_ads_service/pagers.py#L71-L73), where every `SearchPager._method()` call is not retry-decorated.

**Steps to Reproduce:**

Functionally, this occurs with a query returning a large number of results, such that a transient network transport error is likely to occur at least once while paging results. I had a query that had at least 3,260,000 rows (more on this below under "Anything else..."), but this can be observed with any query with more than one page of results.

It is behaviorally evident by putting a breakpoint in `google.api_core.gapic_v1.method._GapicCallable` [where decorators are applied](https://github.com/googleapis/python-api-core/blob/c89f55d10df50c222d05ea046e07e05ef7d17725/google/api_core/gapic_v1/method.py#L137-L141). `retry` is populated with the passed in `Retry()` on the first request, but `retry` is `DEFAULT` for subsequent requests and set to `None`.

**Expected behavior:**

Passing `retry=Retry()` to `GoogleAdsService.search()` calls should apply the retry decoration to every request for each page of results, not just the first request.

**Client library version and API version:**
Client library version: 15.0.0
Google Ads API version: v9 and v10

**Request/Response Logs:**

Can be provided if needed.
 
**Anything else we should know about your project / environment:**

[...]

PR #598, opened 2022-03-15
Retry all search() request pages

Kache:
GoogleAdsService provides `search()`, which accepts a `Retry()`. Apply
the `Retry` decorator on all request pages, not just the first.

Resolves: https://github.com/googleads/google-ads-python/issues/597


BenRKarl:
@Kache thank you for opening this up with these changes! The code you've modified is actually generated, so these changes would be overwritten in future releases. 

To make the changes permanent, we would need for them to get implemented in the ads templates in [gapic-generator-python](https://github.com/googleapis/gapic-generator-python/tree/main/gapic/ads-templates).

Would you be able to file a bug in that library, describing the changes that need to be made? I can help coordinate that as well.

Issue #571, opened 2022-02-06
generate_historical_metrics code sample crashes


jtiscione:
The Python code sample in https://developers.google.com/google-ads/api/docs/keyword-planning/generate-historical-metrics is broken.

It's not actually a code sample, instead it references another code sample https://developers.google.com/google-ads/api/docs/keyword-planning/generate-forecast-metrics and says what needs to be changed. "Make a request to KeywordPlanService.GenerateHistoricalMetrics after creating the KeywordPlan" and then to "Iterate through each of the KeywordPlanHistoricalMetrics and display each of the metrics."

It apparently needs more changes than that, because if I run it, it executes the historical metrics query successfully and I get a response. Then when I try to iterate through the KeywordPlanHistoricalMetrics objects, it reaches the line

`for i, forecast in enumerate(response.keyword_forecasts):`

and it crashes with this error:
python-BaseException
Traceback (most recent call last):
  File "/home/jason/repo/heimdall/.venv/lib/python3.7/site-packages/proto/message.py", line 600, in __getattr__
    pb_type = self._meta.fields[key].pb_type
KeyError: 'keyword_forecasts'

This response object is opaque and I can't tell what's in it at all. It doesn't seem to have a field called "keyword_forecasts".

This would work out better if you actually provided a working code sample for historical metrics instead of referring to the one for forecast metrics, since you're leaving out some of the details needed to get it to work.

PR #619, opened 2022-04-20
Add new generate historical metrics example

BenRKarl:
Based on the request in: #571

### DIFF
--- a/examples/remarketing/add_customer_match_user_list.py
+++ b/examples/remarketing/add_customer_match_user_list.py
@@ -30,19 +30,18 @@ from google.ads.googleads.client import GoogleAdsClient
 from google.ads.googleads.errors import GoogleAdsException
 
 
-def main(client, customer_id, skip_polling):
+def main(client, customer_id):
     """Uses Customer Match to create and add users to a new user list.
 
     Args:
         client: The Google Ads client.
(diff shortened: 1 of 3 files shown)


laurachevalier4 (commented):
Did you mean to include the get_change_details changes here? I thought these were new resource types in v10_1

examples/remarketing/add_customer_match_user_list.py
```diff
@@ -175,22 +173,13 @@ def _add_users_to_customer_match_user_list(
...
-    else:
-        # Wait until the operation has finished.
-        print("Request to execute the added operations started.")
-        print("Waiting until operation completes...")
-        operation_response.result()
-        _print_customer_match_user_list_info(
-            client, customer_id, user_list_resource_name
-        )
-        # [END add_customer_match_user_list]
+    _check_job_status(
```
laurachevalier4, line 176: Can you add a comment about this to match the golden? 
```py
# Offline user data jobs may take 6 hours or more to complete, so instead of waiting for the
# job to complete, retrieves and displays the job status once. If the job is completed
# successfully, prints information about the user list. Otherwise, prints the query to use
# to check the job again later.
```

wihl (approved):

PR #622, opened 2022-04-24
Add url decode for authentication

skudagi:
URL decode enables to parse and generate refresh_token effectively 

Issue solved - when the example is used subsequently, the returning code contains %2F which doesn't get decoded and throws an error

### DIFF
--- a/examples/authentication/authenticate_in_web_application.py
+++ b/examples/authentication/authenticate_in_web_application.py
@@ -37,6 +37,7 @@ import os
 import re
 import socket
 import sys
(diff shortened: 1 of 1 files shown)


bobhancock:
lgtm

bobhancock (approved):
lgtm

BenRKarl:
Hi @skudagi - thanks so much for this! We're going to incorporate this change in our internal release process, where we're making a number of other changes to this example. It should get released in the near future.

PR #667, opened 2022-07-12
fix typo in assignment

belegnar:
Otherwise `AttributeError: Assignment not allowed to repeated field "conversions" in protocol message object.` happens


belegnar:
I've signed CLA, how to restart build?

PR #668, opened 2022-07-13
fix typo in assignment

belegnar:


BenRKarl:
This is technically not a typo as assignment will work just fine, however we tend to use `append` in our other examples so this is a reasonable change.

Issue #662, opened 2022-07-05
Support for proto-plus>=1.19.7


kosteev:
**Describe the problem you are trying to solve:**
Use google-ads in conjunction with other python libraries.

**Describe the solution you'd like:**
Being able to install google-ads latest version together with proto-plus >= 1.19.7 (as it is constraint for some other libraries, e.g. autogenerated google-cloud-datacatalog-lineage which has constraint proto-plus >= 1.19.7)

**Additional context:**
In Cloud Composer we would like to install latest versions of other google libraries like google-cloud-firestore (https://github.com/googleads/google-ads-python/issues/659), google-cloud-datacatalog-lineage, etc., but this is not feasible as google-ads has dependency on old version of proto-plus, and ignoring pip conflicts is not feasible alternative.

It looks like the general concern about adding support for newer versions of proto-plus is that it has performance regression. Is it possible to do performance testing on the fresh versions of this package (https://pypi.org/project/proto-plus/#history) and decide if we can bump constraint to any other >=1.19.7 that doesn't have performance regression?
https://github.com/googleads/google-ads-python/blob/9541a751a2c8d8bd7abf97b0d3a9f61e0ebbcb57/setup.py#L28

This is a severe issue right now for Cloud Composer to have google-ads incompatible with other python google libraries of latest versions.

kosteev:
@bobhancock @BenRKarl I would like to point your attention to this issue as it is severe enough for Cloud Composer team right now.

BenRKarl:
@kosteev thanks for pinging on this issue. We did a round of performance testing in around March that indicated that proto-plus versions newer than `1.19.6` were too slow for this library. We're going to go back and dig into those results, and run some more tests with the most recent versions to see what the trade-offs would be today. In the meantime, is it possible to use older versions of `google-cloud-firestore` and `google-cloud-datacatalog-lineage` that are compatible with `proto-plus 1.19.6`?

BenRKarl:
Thanks @anneadb - we're working on a fix for this that I expect will be part of the next release.

moveingsun:
hi @BenRKarl, is there an estimate time on the next release date? this is also a blocking for us

BenRKarl:
Hi all - we just released version [18.1.0](https://github.com/googleads/google-ads-python/releases/tag/18.1.0), which bumps our dependency on proto-plus to version 1.22.0.

kosteev:
Thank you very much!

moveingsun:
[...]

PR #675, opened 2022-07-27
Increase upper version boundary of proto-plus to 1.20.6

BenRKarl:
Resolves: https://github.com/googleads/google-ads-python/issues/662

After testing newer versions of proto-plus we discovered that they satisfy our benchmarks, so we're opening up the acceptable versions of proto-plus in our library so we're compatible with other Cloud libraries, and to unlock the Composer team.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ install_requires = [
     # have the same version range.
     "grpcio >= 1.38.1, < 2.0.0",
     "grpcio-status >= 1.38.1, < 2.0.0",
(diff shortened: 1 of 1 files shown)


bobhancockg (approved):

jradcliff (approved):

bobhancockg (approved):
lgrm

BenRKarl:
Closing this as the Benchmark results actually did not pass, after further investigation.

PR #691, opened 2022-08-22
Changes for release v11_1.

BenRKarl:


wihl (approved):

AnashOommen (approved):

examples/authentication/generate_user_credentials.py
```diff
@@ -96,7 +96,7 @@ def main(client_secrets_path, scopes):
     )
 
 
-def _get_authorization_code(passthrough_val):
+def get_authorization_code(passthrough_val):
```
AGPapa, line 99: I think there's a mistake here. 

On line 85 in this file this is being called with the old `_get_authorization_code` function name. It's throwing an error when I try to run this example now.

Issue #713, opened 2022-10-23
In compatible requirements with many other packages


UriSShavit:
**Describe the bug:**
I added the usage of the package to my Airflow service.
When trying to compile all the packages dependencies together, using pip-compile (from pip-tools package), there are conflicts with many other packages:
When trying to compile the requirements with google-ads version 14.0.1, the  proto-plus package is locked on version 1.18.1, while all the other packages consist a higher version:
There are incompatible versions in the resolved dependencies:
  proto-plus<2.0.0dev,>=1.22.0 (from google-cloud-memcache==1.4.4->apache-airflow-providers-google==6.8.0->-r requirements.in (line 10))
  proto-plus<2.0.0dev,>=1.22.0 (from google-cloud-pubsub==2.13.10->apache-airflow-providers-google==6.8.0->-r requirements.in (line 10))
  proto-plus<2.0.0dev,>=1.22.0 (from google-cloud-dataproc==5.0.3->apache-airflow-providers-google==6.8.0->-r requirements.in (line 10))
  proto-plus<2.0.0dev,>=1.22.0 (from google-cloud-dataproc-metastore==1.6.3->apache-airflow-providers-google==6.8.0->-r requirements.in (line 10))
  proto-plus<2.0.0dev,>=1.22.0 (from google-cloud-build==3.9.3->apache-airflow-providers-google==6.8.0->-r requirements.in (line 10))
  proto-plus<2.0.0dev,>=1.22.0 (from google-cloud-dataplex==1.1.3->apache-airflow-providers-google==6.8.0->-r requirements.in (line 10))
  proto-plus<2.0.0dev,>=1.22.0 (from google-cloud-redis==2.9.3->apache-airflow-providers-google==6.8.0->-r requirements.in (line 10))
  proto-plus<2.0.0dev,>=1.22.0 (from google-cloud-datacatalog==3.9.3->apache-airflow-providers-google==6.8.0->-r requirements.in (line 10))
  proto-plus<2.0.0dev,>=1.22.0 (from google-cloud-bigquery-datatransfer==3.7.3->apache-airflow-providers-google==6.8.0->-r requirements.in (line 10))
  proto-plus<2.0.0dev,>=1.22.0 (from google-cloud-os-login==2.7.4->apache-airflow-providers-google==6.8.0->-r requirements.in (line 10))
[...]

PR #714, opened 2022-10-23
loosen protobuf requirements to meet standard packages reqs

UriSShavit:
Check out the issue that was raised about that matter: https://github.com/googleads/google-ads-python/issues/713

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ install_requires = [
     "proto-plus == 1.22.0",
     "PyYAML >= 5.1, < 7.0",
     "setuptools >= 40.3.0",
-    "protobuf >= 4.21.5",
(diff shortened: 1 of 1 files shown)


UriSShavit:
@AndrewMBurke can someone please review?

bobhancockg (commented):
Why do we want to do this?

AnashOommen:
@UriSShavit it would be helpful to expand a bit more on *which* standard packages reqs you'd like to meet (names and versions of the packages that interest you). Perhaps this can help us understand the situation better.

bobhancockg:
Airflow has images that support the the supported versions of Google Ads.

BenRKarl:
@UriSShavit just checking whether this PR is still necessary, given the updated Airflow images mentioned in https://github.com/googleads/google-ads-python/issues/713

UriSShavit:
I still think that your requirements should not be that limiting since it breaks all other airflow images in Astronomer and open source version.
I'm going to close this PR as I found a work around (using a lower version of the operator).
I hope that when the time comes and I'll move to a higher Airflow version, I won't encounter this issue again

Issue #735, opened 2023-01-13
Google ADS 19.0.0 dependency tree prevents us to update google-api-core and related modules


Natim:
**Describe the bug:**

When installing google-ads==19.0.0 and then trying to install the last versions of `google-cloud-secret-manager`, `google-cloud-pubsub`, `google-cloud-kms` we have dependencies resolving issues.

**Steps to Reproduce:**

**pyproject.toml**
```
[tool.poetry]
name = "demo"
version = "0.1.0"
description = ""
authors = ["Demo <demo@acme.com>"]

[tool.poetry.dependencies]
python = "~3.10"
google-cloud-pubsub = "2.13.12"
google-cloud-secret-manager = "2.15.0"
google-cloud-storage = "2.7.0"
google-cloud-sqlcommenter = "2.0.0"
google-cloud-kms = "2.14.0"
google-cloud-bigquery = "3.4.1"
google-ads = "19.0.0"
google-api-python-client = "2.72.0"

[build-system]
requires = ["poetry-core>=1.2.0"]
build-backend = "poetry.core.masonry.api"
```

 ```
Because google-api-core (2.11.0) depends on google-api-core (2.11.0)
 and google-api-core (1.34.0) depends on google-api-core (1.34.0), google-api-core (1.34.0 || 2.11.0) requires google-api-core (1.34.0 || 2.11.0).
And because google-ads (19.0.0) depends on google-api-core (2.10.1), google-ads (19.0.0) is incompatible with google-api-core (1.34.0 || 2.11.0).
And because no versions of google-api-core match >1.34.0,<2.0.0 || >2.11.0,<3.0.0dev
 and google-cloud-pubsub (2.13.12) depends on google-api-core (>=1.34.0,<2.0.0 || >=2.11.0,<3.0.0dev), google-ads (19.0.0) is incompatible with google-cloud-pubsub (2.13.12).
So, because demo depends on both google-cloud-pubsub (2.13.12) and google-ads (19.0.0), version solving failed.
```

**Expected behavior:**

We would like `google-ads==19.1.0` to be released to handle the new `protobuf` version and `google-api-core`

**Client library version and API version:**
Client library version: 19.0.0

BenRKarl:
This has been fixed in [version 20.0.0](https://pypi.org/project/google-ads/20.0.0/), released just today.

PR #736, opened 2023-01-13
Allow for a larger dependency scope.

Natim:
Refs #735

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,14 +18,14 @@ import io
 
 install_requires = [
     "google-auth-oauthlib >= 0.3.0, < 1.0.0",
-    "google-api-core == 2.10.1",
+    "google-api-core >= 2.10.1, <= 2.11.0",
     "googleapis-common-protos >= 1.56.4, < 2.0.0",
     # NOTE: Source code for grpcio and grpcio-status exist in the same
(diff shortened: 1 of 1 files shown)


BenRKarl:
@Natim thank you for proposing these changes. It shouldn't be an issue to merge this, we'll just need to run the change through our latency test framework to make sure these new dependencies don't cause performance regressions. I'll do this and report back here.

bobhancockg (approved):

BenRKarl:
@Natim Thanks for following up on this. I had to do a little maintenance on our internal performance framework, but that was fixed as of this morning. I just added this change for testing and should hopefully have enough data on Monday.

Natim:
Hello, we are going back to `google-ads==18.1.0` to not be blocked by this.

GiardEmilien:
We are also stuck by this issue so we cannot use the v12 for now. Do you think this PR will be merged before the sunset date of the v11 (March 31, 2023) ?

BenRKarl:
This change was included in our latest release, [version 20.0.0](https://pypi.org/project/google-ads/20.0.0/), so am closing this out since it's redundant. Thank you for this contribution!

PR #743, opened 2023-01-28
Update setup.py

pymeister:


BenRKarl:
Unfortunately it's not feasible for us to lower the minimum version for protobuf beyond where it already is, so we can't accept this pull request.

PR #752, opened 2023-02-25
Fix bug in normalize_and_hash_email_address function

arnau126:
### First commit

If `email_address` has no `"@"` in it, the function raises `IndexError list index out of range` in:

```
is_gmail = re.match(r"^(gmail|googlemail)\.com$", email_parts[1])
```
because `email_parts` is a list with only one item.

This PR fix this bug by moving the above line inside the if-block `if len(email_parts) > 1` .

### Second commit

I've added `strip` to `email_address` to honor the [Enhanced Conversions doc](https://developers.google.com/google-ads/api/docs/conversions/enhance-conversions#python) which says:

> [...] In order to standardize the hash results, prior to hashing one of these values you must:
> - Remove leading/trailing whitespaces.
> - [...]

### DIFF
--- a/examples/remarketing/upload_enhanced_conversions_for_leads.py
+++ b/examples/remarketing/upload_enhanced_conversions_for_leads.py
@@ -196,7 +196,7 @@ def normalize_and_hash_email_address(email_address):
     Returns:
         A normalized (lowercase, removed whitespace) and SHA-265 hashed string.
     """
-    normalized_email = email_address.lower()
+    normalized_email = email_address.strip().lower()
     email_parts = normalized_email.split("@")
 
     # Check that there are at least two segments
--- a/examples/remarketing/upload_enhanced_conversions_for_leads.py
+++ b/examples/remarketing/upload_enhanced_conversions_for_leads.py
@@ -198,18 +198,17 @@ def normalize_and_hash_email_address(email_address):
     """
     normalized_email = email_address.lower()
     email_parts = normalized_email.split("@")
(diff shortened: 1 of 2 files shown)


arnau126:
About the second commit, I've just realized that the function calls `normalize_and_hash` which already does the strip. However I think that we should keep this extra strip in the very beginning of the function so a string like `"my.email@gmail.com "` (with a trailing whitespace) works.
If we don't perfom this previous strip, the regex won't match, and the period won't be removed.

BenRKarl:
Question - are these changes just to address email address that don't have `@` symbols in them?

arnau126:
@BenRKarl 
Yes.

The original function already tries to do so, but in a wrong way:
```
    email_parts = normalized_email.split("@")
    is_gmail = re.match(r"^(gmail|googlemail)\.com$", email_parts[1])

    # Check that there are at least two segments and the second segment
    # matches the above regex expression validating the email domain name.
    if len(email_parts) > 1 and is_gmail:
```

I've just move the `is_gmail = line` inside the `if` so it's only calculated if there are at least two segments.

BenRKarl:
@arnau126 ok great, sorry for the delay. Looks like these changes are being added to now-deleted files. Could you move them over to the new files where this validation occurs? 

```
remarketing/upload_enhanced_conversions_for_leads.py
remarketing/upload_enhanced_conversions_for_web.py
```

arnau126:
@BenRKarl
Done (and rebased).

BenRKarl (approved):

PR #715, opened 2022-10-24
Adding a full example of how to create a Responsive Search Ad (with keywords and geo targeting)

fblascogarma:
This example shows how to create a complete Responsive Search ad.

Includes creation of: budget, campaign, ad group, ad group ad,
keywords, geo targeting, and image extensions.

PR #753, opened 2023-03-03
Replace main add_responsive_search_ad example with newer version...

BenRKarl:
...that shows how to include geo targeting and image extensions. These changes are simply:

1. Deletes the existing `basic_operations/add_responsive_search_ad.py` example.
2. Moves the `advanced_operations/add_responsive_search_ad_full.py` example so that it's now the main `add_responsive_search_ad.py` example.


AnashOommen (approved):

examples/basic_operations/add_responsive_search_ad.py
```diff
@@ -12,56 +12,270 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""This example adds a responsive search ad to a given ad group.
-
-To get ad groups, run get_ad_groups.py.
 """
+This example shows how to create a complete Responsive Search ad.
```
bobhancock, line 16: What do we mean by complete?

examples/basic_operations/add_responsive_search_ad.py
```diff
@@ -12,56 +12,270 @@
...
+
+Includes creation of: budget, campaign, ad group, ad group ad,
```
bobhancock, line 18: If this is what we mean by complete, consider changing the previous sentence to be 

... to create a Responsive Search Ad with ....

examples/basic_operations/add_responsive_search_ad.py
```diff
@@ -12,56 +12,270 @@
...
+import requests
 
 from google.ads.googleads.client import GoogleAdsClient
 from google.ads.googleads.errors import GoogleAdsException
 
 
-def main(client, customer_id, ad_group_id):
-    ad_group_ad_service = client.get_service("AdGroupAdService")
+# Keywords from user.
+_KEYWORD_TEXT_EXACT_1 = "example of exact match"
```
bobhancock, line 35: The use of an initial underbar is very confusing since variable that begin this way are handled in a specific way by the Python import system. Capital letters are sufficient to indicate global variables.

examples/basic_operations/add_responsive_search_ad.py
```diff
@@ -12,56 +12,270 @@
...
 
-    # Create the ad group ad.
+    ad_group_operation = client.get_type("AdGroupOperation")
+    ad_group = ad_group_operation.create
+    ad_group.name = f"Testing RSA via API {uuid.uuid4()}"
+    ad_group.status = client.enums.AdGroupStatusEnum.ENABLED
+    ad_group.campaign = campaign_resource_name
+    ad_group.type_ = client.enums.AdGroupTypeEnum.SEARCH_STANDARD
+
+    # If you want to set up a max CPC bid uncomment line below.
```
bobhancock, line 212: nit: the line below

examples/basic_operations/add_responsive_search_ad.py
```diff
@@ -77,25 +291,240 @@ def main(client, customer_id, ad_group_id):
...
+    mutate_asset_response = asset_service.mutate_assets(
+        customer_id=customer_id, operations=[asset_operation]
+    )
+    print("Uploaded file(s):")
+    for row in mutate_asset_response.results:
+        print(f"\tResource name: {row.resource_name}")
+
+    image_asset_resource_name = mutate_asset_response.results[0].resource_name
+
+    # Step 6.2 - Create Image Extension
```
bobhancock, line 480: Why are we including this example of extensions? As of 3 April 2023 we want people to stop doing this. Some users, if they are early in the migration process will no longer be able to do this.

I recommend not including any code that deals with extensions.

bobhancock (changes requested):

examples/basic_operations/add_responsive_search_ad.py
BenRKarl: Recall we discussed offline that these changes are just a renaming existing example that we recently approved (https://github.com/googleads/google-ads-python/pull/715).

We'll update this example (along with others) so that it uses assets around when the April 3 migration happens, for now this does illustrate the correct workflow for this type of functionality, right?

BenRKarl:
Closing this PR for now so that we can update the example to use image assets before making it the default example for adding RSAs.

PR #801, opened 2023-08-16
Add an includecode section in get_campaigns.py

AnashOommen:
Add an includecode section in get_campaigns.py so it can be used in the "Getting Started" guide.

### DIFF
--- a/examples/basic_operations/get_campaigns.py
+++ b/examples/basic_operations/get_campaigns.py
@@ -24,7 +24,7 @@ import sys
 from google.ads.googleads.client import GoogleAdsClient
 from google.ads.googleads.errors import GoogleAdsException
 
-
+# [START get_campaigns]
 def main(client, customer_id):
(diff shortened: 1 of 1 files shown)


examples/basic_operations/get_campaigns.py
```diff
@@ -45,6 +45,7 @@ def main(client, customer_id):
                 f'"{row.campaign.name}" was found.'
             )
 
+# [END get_campaigns]
```
BenRKarl, line 48: We usually line these up with the last line in the method, so this should look more like:

```py
    ...
    for batch in stream:
        for row in batch.results:
            print(
                f"Campaign with ID {row.campaign.id} and name "
                f'"{row.campaign.name}" was found.'
            )
            # [END get_campaigns]
```

bobhancockg (approved):

BenRKarl (approved):

PR #815, opened 2023-11-07
Remove the usage of eCPC bidding because of its deprecation

yangda:
To align with the [code example change](https://github.com/googleads/google-ads-java/pull/742) requested for the Java client library.

As seen in this [blog announcement](https://ads-developers.googleblog.com/2023/09/google-ads-shopping-campaign-enhanced.html), any Shopping campaigns using Enhanced cost-per-click (eCPC) will behave as if they are using Manual cost-per-click (CPC) bidding.

Because of the change, I'm updating the examples related to the announcement to reflect the new recommended practice.

### DIFF
--- a/examples/shopping_ads/add_shopping_product_ad.py
+++ b/examples/shopping_ads/add_shopping_product_ad.py
@@ -187,11 +187,15 @@ def add_standard_shopping_campaign(
     # and the ads are ready to serve.
     campaign.status = client.enums.CampaignStatusEnum.PAUSED
 
-    # Sets the bidding strategy to Manual CPC (with eCPC enabled)
+    # Sets the bidding strategy to Manual CPC (with eCPC disabled). eCPC for
+    # standard Shopping campaigns is deprecated. If eCPC is set to true, Google
+    # Ads ignores the setting and behaves as if the setting was false. See this
+    # blog post for more information:
+    # https://ads-developers.googleblog.com/2023/09/google-ads-shopping-campaign-enhanced.html
     # Recommendation: Use one of the automated bidding strategies for Shopping
     # campaigns to help you optimize your advertising spend. More information
     # can be found here: https://support.google.com/google-ads/answer/6309029
-    campaign.manual_cpc.enhanced_cpc_enabled = True
+    campaign.manual_cpc.enhanced_cpc_enabled = False
 
     # Sets the budget.
     campaign.campaign_budget = budget_resource_name


examples/shopping_ads/add_shopping_product_ad.py
```diff
@@ -187,11 +187,15 @@ def add_standard_shopping_campaign(
...
+    # Sets the bidding strategy to Manual CPC (with eCPC disabled). eCPC for
+    # standard Shopping campaigns is deprecated. If eCPC is set to true, Google
+    # Ads ignores the setting and behaves as if the setting was false. See this
+    # blog post for more information:
+    # https://ads-developers.googleblog.com/2023/09/google-ads-shopping-campaign-enhanced.html
     # Recommendation: Use one of the automated bidding strategies for Shopping
     # campaigns to help you optimize your advertising spend. More information
     # can be found here: https://support.google.com/google-ads/answer/6309029
-    campaign.manual_cpc.enhanced_cpc_enabled = True
+    campaign.manual_cpc.enhanced_cpc_enabled = False
```
BenRKarl, line 198: Wouldn't it be better to not set this field at all, since the default value would be false? This line might give the impression that eCPC needs to be manually disabled, instead of being disabled by default.

yangda: Thanks, Ben! Josh raised the same question for the Java pull request. It seems that the default value is still true if unset based on my test (I was using `client.copy_from(campaign.manual_cpc, client.get_type("ManualCpc"))`). The deprecation was a backend change that did not affect the API's behavior (i.e. if previously unset => true, now it behaves the same).

BenRKarl: I tested this in the example in both ways,  by not setting the field, and by setting `manual_cpc` in the way that you demonstrated, (`campaign.manual_cpc.enhanced_cpc_enabled`) and in both cases the field evaluates to `False`. Perhaps this is specific to Python, but I don't think that's true because I know that the default value for boolean fields is false at the protobuf level ([link](campaign.manual_cpc.enhanced_cpc_enabled)). Given this I still think it makes more sense to not set this field, instead of explicitly setting it to false.

jradcliff: See discussion in the [Java PR](https://github.com/googleads/google-ads-java/pull/742#discussion_r1370570124).

BenRKarl: Discussed with Josh offline and I'm OK with this change, however still a little confused about the API behavior that we can take offline.

BenRKarl (approved):

bobhancockg (approved):

PR #829, opened 2024-01-11
Demonstrate passing consent in offline conversion upload

jradcliff:
Ref: b/310739611

Change-Id: Ide8079c878fb3f5a5d5d7220e0e00a35d987d4f3


bobhancockg (approved):
Since we require the user to supply up to 11 command line arguments, we should consider providing the option of supplying a TOML file.

BenRKarl (approved):
I'm assuming that adding consent is not always required, otherwise the input argument would be required.

PR #883, opened 2024-08-28
Add example code for PMax campaigns to add search theme signal

yangda:
**Context**: currently the `add_performance_max_campaign.py` example only contains code samples to add the audience signal as an asset group signal. We also support adding search themes as an asset group signal, which is documented [here](https://developers.google.com/google-ads/api/performance-max/asset-group-signals#search_themes). With that we want to modify our code example to include it.

**Changes**:

1. Added code to demonstrate adding a search theme asset group signal.
2. Updated the document links in comments.
3. Updated the positions of utility comments (`[START/END add_performance_max_campaign_*]`) to prepare for devsite updates.

### DIFF
--- a/examples/advanced_operations/add_performance_max_campaign.py
+++ b/examples/advanced_operations/add_performance_max_campaign.py
@@ -120,6 +120,9 @@ def main(client, customer_id, audience_id):
         headline_asset_resource_names,
         description_asset_resource_names,
     )
+    asset_group_signal_operations = create_asset_group_signal_operations(
+        client, customer_id, audience_id
+    )
 
     mutate_operations = [
         # It's important to create these entities in this order because
@@ -130,15 +133,8 @@ def main(client, customer_id, audience_id):
         # other mutate operations
         *campaign_criterion_operations,
         *asset_group_operations,
+        *asset_group_signal_operations,
     ]
-
-    # Append an asset group signal operation is an audience ID is given.
-    if audience_id:
-        mutate_operations.append(
-            create_asset_group_signal_operation(
-                client, customer_id, audience_id
-            )
-        )
 
     # Send the operations in a single Mutate request.
     response = googleads_service.mutate(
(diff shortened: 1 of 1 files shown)


examples/advanced_operations/add_performance_max_campaign.py
```diff
@@ -120,6 +120,9 @@ def main(client, customer_id, audience_id):
         headline_asset_resource_names,
         description_asset_resource_names,
     )
+    asset_group_signal_operations = create_asset_group_signal_operations(
+        client, customer_id, audience_id
```
devchas, line 124: Does this work if no audience_id is provided or will it throw an error?

yangda: With my test it throws the following error when `audience_id` is `None`:
```
[2024-08-29 11:43:41 - WARNING] Request made: ClientCustomerId: REDACTED, Host: googleads.googleapis.com, Method: /google.ads.googleads.v17.services.GoogleAdsService/Mutate, RequestId: REDACTED, IsFault: True, FaultMessage: 'None' part of the resource name is invalid.
Request with ID "REDACTED" failed with status "NOT_FOUND" and includes the following errors:
Error with message "'None' part of the resource name is invalid.".
                On field: mutate_operations
                On field: asset_group_signal_operation
                On field: create
                On field: audience
                On field: audience
```

devchas: Can you adjust the behavior so that you skip that part of the code if audience ID isn't provided, similar to how the example worked before?

yangda: The `audience_id` check logic was moved to L650 below, within the function `create_asset_group_signal_operations`.

I think I misunderstood what you meant in the previous comment. The test was done by removing the `if audience_id:` check at L650 to observe what happens if `audience_id` is `None`. Running the complete updated example works without issues.

BenRKarl (approved):

Issue #896, opened 2024-11-12
Add support for Python 3.13


Dumeng:
**Describe the problem you are trying to solve:**
The lib currently has a Python version requirement limit on < 3.13 only. And this causes a pip install error on the stable Python release.
https://github.com/googleads/google-ads-python/blob/6c205c4abdf95a18c70b938e37f448660b280da9/setup.py#L55

**Describe the solution you'd like:**
Lift the Python version requirement.

**Describe alternatives you've considered:**
We want to leverage the feature in the latest Python stable version. So, it looks like there are no actual alternatives.

**Additional context:**

evil-shrike:
Besides the support itself (which is a must), currently all READMEs (on GitHub and PyPi) advertise the 3.8+ requirement only, not mentioning that it's actually <3.13 , so the last package that will install on 3.13 is 22.0.0 which super weird. 

I don't get in general the need for adding "<3.13" requirement.

mmierzwa:
Is the update to come any soon?

perfa-funnel:
While I have created a PR for this, I have found that the `--ignore-requires-python` flag for `pip` installs this without issue and it seems to work fine.

GAC-BSP:
> `--ignore-requires-python`

Is anybody aware of a workaround when using poetry + pyproject.toml?

BenRKarl:
Hi all, 3.13 support was added in the latest release, 26.0.0.

PR #903, opened 2025-01-28
Add support for python 3.13 (#896)

perfa-funnel:
Remove python 3.13 upper bound, and adjust version requirements for `grpcio` and `grpcio-status` to support this change.

Technically they don't need to be bumped, but are not available as cheese wheels for Python 3.13, so require build dependencies in the test environment or a cached wheel built manually.

Tested in the Docker image with the `nox` command.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,8 @@ RUN apt-get update -qy && \
         python3.9 \
         python3.10 \
         python3.11 \
-        python3.12 && \
+        python3.12 \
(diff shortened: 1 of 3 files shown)


funnel-tobias (approved):
👍

mmierzwa:
Any chance to review and merge it soon?

BenRKarl:
Thanks for the PR! Support for 3.13 was added in a separate PR and is available in version 26.0.0.

PR #911, opened 2025-02-17
python3.13 support

hamedsh:
add python 3.13 support


BenRKarl:
Thank you for the PR! This was added in a separate PR and is available in version 26.0.0.

PR #912, opened 2025-02-27
Changes for release v19.

BenRKarl:


AnashOommen (approved):

examples/advanced_operations/add_performance_max_campaign.py
```diff
@@ -104,11 +106,10 @@ def main(client, customer_id, audience_id):
...
     )
-    performance_max_campaign_operation = (
-        create_performance_max_campaign_operation(
-            client,
-            customer_id,
-        )
+    performance_max_campaign_operation = create_performance_max_campaign_operation(
+        client,
+        customer_id,
+        brand_guidelines_enabled,
```
bobhancockg, line 112: Although it should not affect execution since a function signature is a tuple, This occurs in a couple of places. I had asked him to remove the comma after the last argument so it is 'def foo(a,b,c)' and not 'def foo(a,b,c,)'. We should make that change since it makes the arguments look unfinished. This won't block release.

bobhancockg (changes requested):
One formatting issue we should fix if we can, but it won't block release.

bobhancockg:
lgtm

bobhancockg (approved):

Issue #917, opened 2025-03-09
Unable to clone the repository on Windows due to invalid file path


PragyanTiwari:
**Describe the bug:**

The Directory i.e. : [google-ads-python](https://github.com/googleads/google-ads-python/tree/main)/[tests](https://github.com/googleads/google-ads-python/tree/main/tests)/[constraints](https://github.com/googleads/google-ads-python/tree/main/tests/constraints)/minimums/
containing the file `constraints-<3.13.txt` consists an invalid character `<`, causing to successfully clone the repo but failed to checkout.

**Steps to Reproduce:**
running this on Win11 (powershell terminal):

```
git clone https://github.com/googleads/google-ads-python.git google-ads-api-python-client
```

**Ouput**
```
Cloning into 'google-ads-api-python-client'...
remote: Enumerating objects: 53265, done.
remote: Counting objects: 100% (16331/16331), done.
remote: Compressing objects: 100% (3481/3481), done.
remote: Total 53265 (delta 14635), reused 14195 (delta 12843), pack-reused 36934 (from 2)
Receiving objects: 100% (53265/53265), 21.29 MiB | 3.45 MiB/s, done.
Resolving deltas: 100% (44621/44621), done.
error: invalid path 'tests/constraints/minimums/constraints-<3.13.txt'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'
```

**Suggestion:**
*Rename the file and modify the file paths in the unit tests.*

**Client library version and API version:**
Client library version: 26.0.0
Google Ads API version: v15_0
 
**Anything else we should know about your project / environment:**
git version: 2.48.1.windows.1

PR #920, opened 2025-03-10
File renamed to resolve constraints and modified noxfile.py

PragyanTiwari:
# Pull request regarding issue: #917 

**key changes:**
---
- renamed `constraints-<3.13.txt` to `constraints-prev-3.13.txt` , removing the `<` character.
- updated `noxfile.py` for running unit tests correctly.


BenRKarl (approved):

noxfile.py
```diff
@@ -75,7 +75,7 @@ def tests_minimum_dependency_versions(session, protobuf_implementation):
         # If runnning a Python version other than 3.13 then use the constraints
         # file intended for all versions previous to 3.13.
         # TODO: Update this when new major versions of Python are adopted.
-        filename = "constraints-<3.13.txt"
```
BenRKarl, line 78: Instead of `prev` could you use `lt` for "less than" - "previous" implies that it's defining the version previous to 3.13, whereas "less than" implies any version that's less than 3.13.

PragyanTiwari: How about `less-than` instead of `lt` ? if name is not an important constraint here. Since, `less-than` fits better for the interpretation.
If this isn't so, then I'll go with your changes.

BenRKarl: `less-than` works great, if you can make that update I'll merge this.

BenRKarl (approved):

PR #922, opened 2025-03-11
Flesh out import statements in vXX/services/__init__.py 

BenRKarl:
So types defined under services are accessible just like all other types.


BenRKarl:
> This looks good. Are we generating this automatically?

This change was manual, but I'm working with the GAPIC team to build this into the templates, which would make it automatic.

bobhancockg (approved):

PR #934, opened 2025-05-09
Add demand gen campaign example.

BenRKarl:


examples/advanced_operations/add_demand_gen_campaign.py
```diff
@@ -0,0 +1,412 @@
+# Copyright 2024 Google LLC
```
Raibaz, line 1: 2025 :)

Raibaz (dismissed):
One small nit, otherwise LGTM

examples/advanced_operations/add_demand_gen_campaign.py
```diff
@@ -0,0 +1,412 @@
...
+    mutate_operation = client.get_type("MutateOperation")
+    asset_operation = mutate_operation.asset_operation
+    asset = asset_operation.create
+    asset.resource_name = asset_resource_name
+    # Provide a unique friendly name to identify your asset. When there is an
+    # existing image asset with the same content but a different name, the new
+    # name will be dropped silently.
+    asset.name = asset_name
+    asset.type_ = client.enums.AssetTypeEnum.IMAGE
+    asset.image_asset.data = requests.get(url).content
```
bobhancockg, line 344: Consider calling this outside of this function. The function should be to create the asset only. Separating the retrieval of the data from the creation of the asset separate error processing for network retrieval from the asset creation process. Also, if, as in this example, it is called repeatedly from a list comprehension it avoid redundant network calls if the same data is used.

There should be error handling around all network calls. i.e., a try except block.

examples/advanced_operations/add_demand_gen_campaign.py
```diff
@@ -0,0 +1,412 @@
...
+    ad.final_urls.append(DEFAULT_FINAL_URL)
+
+    demand_gen_ad = ad.demand_gen_video_responsive_ad
+    # Ensure business_name is an AssetLink and assign text to its text_asset.text
+    demand_gen_ad.business_name.text = "Interplanetary Cruises"
+    # If it needs to be an AssetLink to a text asset,
+    # that would require creating another asset.
+
+    # Create AssetLink for video
+    video_asset_link = client.get_type("AdVideoAsset")
```
bobhancockg, line 293: Consider making these all single line operations with keywords. It increases readability and makes the process clearer. I think this syntax should work, but verify it. The order of operations, left to right, should return the link.asset object first and then use the tuple as the asset argument.

In general, in the examples, we should try to make these type of operations into a single statement if possible.

 demand_gen_ad.videos.append(
    client.get_type("AdVideoAsset")(asset=video_asset_resource_name)
)

demand_gen_ad.logo_images.append(
         client.get_type("AdImageAsset")(asset=logo_asset_resource_name)
)
etc.

N.B. Comments in github do not respect formatting. So, not correct indentation.

examples/advanced_operations/add_demand_gen_campaign.py
```diff
@@ -0,0 +1,412 @@
...
+    args = parser.parse_args()
+
+    # GoogleAdsClient will read the google-ads.yaml configuration file in the
+    # home directory if none is specified.
+    googleads_client = GoogleAdsClient.load_from_storage(version="v19")
+
+    try:
+        main(googleads_client, args.customer_id, args.video_id)
+    except GoogleAdsException as ex:
+        print(
```
bobhancockg, line 403: Since a lot of the feedback that we have received from HaTS is that beginning users would like to have plug and play examples, it may be worthwhile to mention that in a production application the print statements should be replaced with logging statements.

An overall change for all examples would be to remove the print statements and replace them with real logging. This will demonstrate to users how to implement logging with error handling. We can include a blurb in the README on how to monitor logging files real time, like 'tail -f log_file` in Linux or Mac OS. We would just need to research how to do it in Windows.

examples/advanced_operations/add_demand_gen_campaign.py
BenRKarl: Done

examples/advanced_operations/add_demand_gen_campaign.py
BenRKarl: This style won't work as-is:

```
client.get_type("AdVideoAsset")(asset=video_asset_resource_name)
```

In `get_type` the message class is instantiated, so what's returned isn't callable. That line raises a `object is not callable` error.

examples/advanced_operations/add_demand_gen_campaign.py
BenRKarl: Since we do this in a few examples, I moved this to a separate utility function with exception handling.

Raibaz (approved):

PR #946, opened 2025-05-24
Add type hints to examples/feeds

bobhancockg:
This change adds type hints to function signatures and variable annotations to the Python files in the examples/feeds directory. This improves code readability and helps with static analysis.


BenRKarl:
Closing because examples/feeds should no longer exist.

PR #947, opened 2025-05-24
Add type hints to analyzer.py

bobhancockg:
This change adds type annotations and necessary imports to the examples/google-ads-account-analyzer-demo/analyzer.py script.

Type hints were added to the following functions:
- account_hierarchy_module
- print_account_hierarchy
- get_users_module

I verified the annotations, and one issue related to an untyped dictionary was fixed. Remaining errors originate from the google-ads library itself.


sundquist (dismissed):
LGTM

PR #953, opened 2025-05-24
Added type hints to examples/remarketing

bobhancockg:
Add type hints to files in examples/remarketing

This commit adds type annotations to Python files in the examples/remarketing directory.

The following files were updated:
- examples/remarketing/__init__.py (no changes as it was empty)
- examples/remarketing/add_conversion_action.py
- examples/remarketing/add_conversion_based_user_list.py
- examples/remarketing/add_custom_audience.py
- examples/remarketing/add_customer_match_user_list.py
- examples/remarketing/add_dynamic_remarketing_asset.py
- examples/remarketing/add_flexible_rule_user_list.py
- examples/remarketing/add_logical_user_list.py
- examples/remarketing/add_merchant_center_dynamic_remarketing_campaign.py

Type hints were added to functions, arguments, and variables in these files. Necessary imports from the `typing` module and the `google.ads.googleads` library were also added. The changes ensure the code remains functionally identical.

This addresses the issue of adding annotations and type hints to these files. The remaining files in the directory will be addressed in subsequent commits.


examples/remarketing/add_customer_match_user_list.py
```diff
@@ -571,11 +630,19 @@ def normalize_and_hash(s, remove_all_whitespace):
...
         ),
     )
+    # Need to instantiate a GoogleAdsClient to access enums, but this is only
+    # used for help text, so it's fine if it fails.
+    try:
+        consent_status_enum_names = [
+            e.name for e in GoogleAdsClient.load_from_storage(version="v19").enums.ConsentStatusEnum
+        ]
+    except Exception:
+        consent_status_enum_names = ["UNSPECIFIED", "UNKNOWN", "GRANTED", "DENIED"]
```
sundquist, line 640: My understanding is that this would be used for the help text in the case of an exception. If my understanding is correct about this being help text, then only GRANTED and DENIED are recommended values to set for consent. UNKNOWN is output-only, so you'll get an error if setting to UNKNOWN. You can use UNSPECIFIED, but it's like setting the value to empty, and it has no effect.

sundquist (changes requested):

BenRKarl (dismissed):

examples/remarketing/add_customer_match_user_list.py
BenRKarl: I updated so that `UNSPECIFIED` and `UKNOWN` are filtered out when the options are listed.

sundquist (approved):
LGTM

PR #995, opened 2025-05-31
Gold add error handling tests

bobhancockg:
Conclusion of Debugging Efforts for `deserialize`:
The core issue is mocking the `deserialize` static method, which is called on a class that is dynamically obtained in the production code via `type(client.get_type("GoogleAdsFailure"))`.
- The `__class__` assignment strategy did not yield the expected behavior for `type()` lookup on the mock instance.
- Direct patching attempts using `@patch` have been problematic due to difficulties in finding the correct, resolvable string path for the dynamically determined type, culminating in the current `ModuleNotFoundError`.

Further work is needed to find a robust mocking strategy for this specific scenario. Ideally, the code in `test_handle_partial_failure.py` should be reverted to use the `__class__` assignment strategy to resolve the `ModuleNotFoundError` and go back to the `AttributeError`, which is a less severe failure mode.

### DIFF
--- a/examples/error_handling/tests/.keep
+++ b/examples/error_handling/tests/.keep
@@ -1,0 +1,1 @@
+# This is a placeholder file to ensure the directory is tracked by git.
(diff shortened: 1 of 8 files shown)


BenRKarl:

PR #1004, opened 2025-06-02
Gold tests for utils

bobhancockg:
…v19 compatibility

This commit introduces a test suite for the utility modules located in the `examples/utils/` directory.

Key changes:
- Created a new test directory: `examples/utils/tests/`.
- Added `examples/utils/tests/__init__.py`.
- Added `examples/utils/tests/test_example_helpers.py` with unit tests for `get_printable_datetime` and `get_image_bytes_from_url` using the `unittest` framework and `unittest.mock`.
- Verified that the existing `google-ads` client library (v26.1.0) already supports Google Ads API v19, so no direct dependency changes for API versioning were required.
- Updated `noxfile.py`: Modified the `TEST_COMMAND` for `unittest discover` to scan from the project root (`-s .`) instead of just the top-level `tests` directory. This ensures the new tests in `examples/utils/tests/` are discovered and executed by the standard `nox` testing sessions.
- Adjusted the import statement in `test_example_helpers.py` to `from examples.utils import example_helpers` to ensure correct module resolution when tests are run from the project root.

### DIFF
--- a/examples/utils/tests/test_example_helpers.py
+++ b/examples/utils/tests/test_example_helpers.py
@@ -1,0 +1,89 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


BenRKarl:

PR #1050, opened 2025-12-08
Fix example to also normalize plus character in emails

lucasra1:
Correct function to follow documentation. `+` characters also need to be removed and normalization is not only for `gmail.com` or `googlemail.com` addresses

https://developers.google.com/google-ads/api/docs/conversions/upload-offline#prepare-data
> For email addresses:
>  - Remove all periods (.) in the username (before the @ symbol). For example,   
>     jane.doe@example.com becomes janedoe@example.com.
>   - Remove the plus (+) symbol and any characters following it in the username (before the @ 
>     symbol). For example, janedoe+newsletter@example.com becomes janedoe@example.com.

### DIFF
--- a/examples/remarketing/upload_enhanced_conversions_for_leads.py
+++ b/examples/remarketing/upload_enhanced_conversions_for_leads.py
@@ -254,14 +254,10 @@ def normalize_and_hash_email_address(email_address: str) -> str:
 
     # Check that there are at least two segments
     if len(email_parts) > 1:
-        # Checks whether the domain of the email address is either "gmail.com"
-        # or "googlemail.com". If this regex does not match then this statement
-        # will evaluate to None.
-        if re.match(r"^(gmail|googlemail)\.com$", email_parts[1]):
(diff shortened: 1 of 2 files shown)


lucasra1:
> Thank you for submitting this PR! I'm conferring internally on some of the details here to make sure I understand the functionality. In the meantime can you look at the CLA mentioned [here](https://github.com/googleads/google-ads-python/pull/1050#issuecomment-3629212238)?


If I am not mistaken I have agreed to it and rerun the GitHub Action successfully

bobhancockg (changes requested):

bobhancockg (approved):

BenRKarl (approved):

grachov:
@bobhancockg @BenRKarl Is the change in the PR correct? It currently removes only `+`, but not the postfix coming after it, so `janedoe+newsletter@example.com` becomes `janedoenewsletter@example.com` and not `janedoe@example.com` - making it completely different email.

PR #1066, opened 2026-04-01
fix: await response.code() and use sys.exc_info() in async gRPC interceptor

danishashko:
Fixes #1059

Two bugs in `_AsyncExceptionInterceptor._handle_grpc_failure_async()`:

**Bug 1: `response.exception()` raises `AttributeError`**

`grpc.aio.Call` objects don't have an `.exception()` method. That method only exists on synchronous `grpc.Future` objects. When a timeout or temporary service error triggers the async interceptor, calling `response.exception()` raises `AttributeError: 'UnaryUnaryCall' object has no attribute 'exception'`, masking the original gRPC error entirely.

Fix: use `sys.exc_info()[1]` to capture the live exception — this works since `_handle_grpc_failure_async` is always called from within an `except grpc.RpcError` block in the async wrappers.

**Bug 2: `response.code()` not awaited**

In `grpc.aio`, `Call.code()` is a coroutine and must be awaited. Without `await`, `status_code` would be a coroutine object, making the `status_code not in _RETRY_STATUS_CODES` check always evaluate to `True`, meaning retry-able errors (INTERNAL, RESOURCE_EXHAUSTED) would incorrectly be parsed as GoogleAdsFailures instead of being re-raised as-is.

**Other cleanup:**
- Removed leftover reasoning comments that were accidentally committed
- Import `_RETRY_STATUS_CODES` from `interceptor.py` instead of redefining it locally

### DIFF
--- a/google/ads/googleads/interceptors/exception_interceptor.py
+++ b/google/ads/googleads/interceptors/exception_interceptor.py
@@ -26,7 +26,10 @@ import grpc
 
 from google.ads.googleads import util
 from google.ads.googleads.errors import GoogleAdsException
+import sys
+
 from google.ads.googleads.interceptors import Interceptor, ContinuationType
+from google.ads.googleads.interceptors.interceptor import _RETRY_STATUS_CODES
 from google.ads.googleads.interceptors.response_wrappers import _UnaryStreamWrapper, _UnaryUnaryWrapper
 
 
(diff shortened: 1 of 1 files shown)


bobhancockg (approved):

ZLeventer (commented):
Good fix. Two real bugs here:

1. **`await response.code()`** — `grpc.aio.Call.code()` is a coroutine, unlike the sync `grpc.Future.code()` which returns immediately. Calling it without `await` would return a coroutine object, not the status code, causing the `not in _RETRY_STATUS_CODES` check to always be `True` (a coroutine is never equal to a status code enum).

2. **`sys.exc_info()[1]` instead of `response.exception()`** — `grpc.aio.Call` doesn't expose `.exception()` the same way sync `grpc.Future` does. Using `sys.exc_info()` within the `except` block is the correct approach since this method is always called from an exception handler.

Also good cleanup: removing the commented-out reasoning/notes and the unreachable `raise response.exception()` at the bottom. The import of `_RETRY_STATUS_CODES` from the module instead of redefining it locally is the right call — DRY and ensures any future changes to the retry set apply everywhere.

BenRKarl:
@danishashko could you complete the required CLA? See [this comment](https://github.com/googleads/google-ads-python/pull/1066#issuecomment-4170391147).

PR #1074, opened 2026-05-13
Update experiments examples per required usage flows.

laurachevalier4:
Change-Id: Ifba0c463bfc763cc2a9032fc800720c67f304261

### DIFF
--- a/examples/experiments/create_asset_optimization_experiment.py
+++ b/examples/experiments/create_asset_optimization_experiment.py
@@ -25,25 +25,6 @@ from uuid import uuid4
 from examples.utils.example_helpers import get_image_bytes_from_url
 from google.ads.googleads.client import GoogleAdsClient
 from google.ads.googleads.errors import GoogleAdsException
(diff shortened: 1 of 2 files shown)


examples/experiments/create_asset_optimization_experiment.py
```diff
@@ -42,6 +42,24 @@ def main(
...
+        customer_id=customer_id, query=query
+    )
+    campaign_resource_name = None
+    for row in search_response:
+        campaign_resource_name = row.asset_group.campaign
+        break
+
+    if not campaign_resource_name:
+        print(f"Asset group with ID {asset_group_id} not found.")
+        sys.exit(1)
```
BenRKarl, line 58: Since this query can only return a single asset_group, it's better not to initialize a loop and instead to just index directly to get the first item:

```py
if len(search_response):
    campaign_resource_name = search_response[0].asset_group.campaign
else:
    print(f"Asset group with ID {asset_group_id} not found.")
    sys.exit(1)
```

BenRKarl (approved):

PR #1075, opened 2026-05-19
change the version tag

asmaT2023:


BenRKarl:
This would revert the library version back by five major versions. I don't understand the need for this change, could you provide more information as to why it's necessary?

